PR #1574, opened 2020-05-10
Enable/Disable headers table actions

mahdibenromdhane:


LucasBrazi06:
> The right coding to deal with disable/enable button is this one (otherwise you have a bug):
> `this.removeAction.disabled = (assets.count === 0);`
> And the code seems not in sync with master-qa.
> Double check also that all the files containing an Add/Remove button has the same coding.

You can ignore this message, believed that this issue was in the To Be Reviewed column ;-)
But double check that all the files containing an Add/Remove button has the same coding.
Thanks!

mahdibenromdhane:
> > The right coding to deal with disable/enable button is this one (otherwise you have a bug):
> > `this.removeAction.disabled = (assets.count === 0);`
> > And the code seems not in sync with master-qa.
> > Double check also that all the files containing an Add/Remove button has the same coding.
> 
> You can ignore this message, believed that this issue was in the To Be Reviewed column ;-)
> But double check that all the files containing an Add/Remove button has the same coding.
> Thanks!

It's missing in user-sites-table-data-source.ts, I will handle it in my next PR

LucasBrazi06:
> It's missing in user-sites-table-data-source.ts, I will handle it in my next PR
If it's only one file I will do it ;-)
Thanks!

PR #1590, opened 2020-05-13
Add default car and car type to car create form

mahdibenromdhane:


src/app/pages/cars/car/car.component.html
```diff
@@ -52,6 +52,24 @@
...
               </mat-form-field>
             </div>
+            <div class="form-group" *ngIf="isBasic">
+              <div style="text-align: left">
+              <mat-checkbox [formControl]="isDefault">
+                {{'cars.default_car' | translate}}
+              </mat-checkbox>
+            </div>
+            </div>
+            <div class="form-group" *ngIf="isBasic">
```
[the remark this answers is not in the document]
mahdibenromdhane: For this PR we are handling only the basic user. 
The admin case will be handled later.

src/app/pages/cars/car/car.component.ts
```diff
@@ -72,13 +74,23 @@ export class CarComponent implements OnInit {
...
+          Validators.required,
+        ]))
     });
     // Form
     this.id = this.formGroup.controls['id'];
     this.vin = this.formGroup.controls['vin'];
     this.licensePlate = this.formGroup.controls['licensePlate'];
     this.carCatalogID = this.formGroup.controls['carCatalogID'];
     this.carCatalog = this.formGroup.controls['carCatalog'];
+    this.carCatalog = this.formGroup.controls['carCatalog'];
```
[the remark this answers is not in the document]
mahdibenromdhane: Done

src/app/pages/cars/car/car.component.html
```diff
@@ -52,6 +52,24 @@
                 </mat-error>
               </mat-form-field>
             </div>
+            <div class="form-group" *ngIf="isBasic">
+              <div style="text-align: left">
+              <mat-checkbox [formControl]="isDefault">
```
OliveGerste, line 57: I think you have to add a default value here.
When you create a car and do not touch the checkbox the request looks like this:
_id: ""
vin: "W0L000051T2123456"
licensePlate: "PrivateNoDefault"
carCatalogID: "1271"
carCatalog: "Audi e-tron S"
isDefault: ""
isPrivate: "true"_

And the created entry like this:
image

OliveGerste: Like I mentioned in the backend PR. 
The request will not work anymore if you want to add a pool car, because you can just say if it is a private car or not and not f it is a company car or a pool car. 
Maybe you should take the type of the car in the request and not whether it is private or not.

mahdibenromdhane: > Like I mentioned in the backend PR.
> The request will not work anymore if you want to add a pool car, because you can just say if it is a private car or not and not f it is a company car or a pool car.
> Maybe you should take the type of the car in the request and not whether it is private or not.

Yes true !
done

PR #1613, opened 2020-05-17
Comment out manual refresh mechanism.

jerome-benoit:
In preparation for a removal.

Signed-off-by: Jérôme Benoit <jerome.benoit@piment-noir.org>

### DIFF
--- a/src/app/pages/charging-stations/charging-limit/charging-profile-limit/charging-station-charging-profile-limit-schedule-table-data-source.ts
+++ b/src/app/pages/charging-stations/charging-limit/charging-profile-limit/charging-station-charging-profile-limit-schedule-table-data-source.ts
@@ -93,10 +93,9 @@ export class ChargingStationChargingProfileLimitScheduleTableDataSource extends 
 
   public setChargingProfileSchedule(schedules: Schedule[]) {
     this.schedules = schedules;
-    this.getManualDataChangeSubject().next();
+    // this.getManualDataChangeSubject().next();
   }
   public setCharger(charger: ChargingStation) {
     this.charger = charger;
   }
-
 }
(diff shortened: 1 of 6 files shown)


LucasBrazi06:
It should be still configurable, if we set socketIO to false, pooling every 'n' secs should work again.
This manual refresh is only for Editable datasource.
Just double check.

jerome-benoit:
> It should be still configurable, if we set socketIO to false, pooling every 'n' secs should work again.
> This manual refresh is only for Editable datasource.
> Just double check.

The polling will still work without that code but the display refresh will not happen right away but since only cars and charging profiles list is using it, removing it will have no real impact on the dashboard. 

How about having to maintain only one display refreshing mechanism: 

+ Remove pooling: not real time, need to maintain other observable lists for displaying real time changes (that nobody is using)
+ Use socket.io DB changes pushing: real time, observable lists are the same everywhere, can be fine grained. 

What do you think ?

LucasBrazi06:
> How about having to maintain only one display refreshing mechanism:
> 
> * Remove pooling: not real time, need to maintain other observable lists for displaying real time changes (that nobody is using)
> * Use socket.io DB changes pushing: real time, observable lists are the same everywhere, can be fine grained.
> 
> What do you think ?

I call you :-).

src/app/pages/charging-stations/charging-limit/charging-profile-limit/charging-station-charging-profile-limit-schedule-table-data-source.ts
```diff
@@ -93,10 +93,9 @@ export class ChargingStationChargingProfileLimitScheduleTableDataSource extends
 
   public setChargingProfileSchedule(schedules: Schedule[]) {
     this.schedules = schedules;
-    this.getManualDataChangeSubject().next();
```
LucasBrazi06, line 96: The list of charging profile is no longer displayed.
I'll put it back.

jerome-benoit: It's not refreshed because socket.io is not listening to changes on the right collection yet in that PR.

LucasBrazi06: > It's not refreshed because socket.io is not listening to changes on the right collection yet in that PR.

No, the call to the DB is already done, these data sources are disconnected from the DB thus the 'manual refresh' when assigned new values.

PR #1633, opened 2020-05-23
Admin Create/edit car, assign/unassign User 

mahdibenromdhane:


OliveGerste:
I tested again and found some small issues:

- [ ] When you open the converter dialog twice, every connector is selected by default:
image

- [ ] A user can have multiple default cars by clicking the checkbox in each dialog for each car.

- [ ] When changing the car owner. The message says 0 users have been updated.
image

- [ ] The car owner property is not saved/loaded correctly. (when changing just this property)

LucasBrazi06:
@fcoutellier relevant for translation once merged

PR #1708, opened 2020-06-12
Add and enable spanish translation 

jerome-benoit:


LucasBrazi06:
@jerome-benoit New languages not available in user's language drop down

LucasBrazi06:
@jerome-benoit Thanks by the way for the integration!

jerome-benoit:
> @jerome-benoit New languages not available in user's language drop down

Configuration changes are now needed in all landscape to add es_MX. You can test locally by doing the same changes done in configuration template files.

PR #1768, opened 2020-06-27
socket.io: debug jwt token expiration.

jerome-benoit:


src/app/utils/SocketIOClient.ts
```diff
@@ -35,6 +35,12 @@ export default class SocketIOClient {
     } else {
       // console.log('Missing serverURL and token arguments');
     }
+    this.socketIO.on('unauthorized', (error) => {
+      if (error.data.type === 'UnauthorizedError' || error.data.code === 'invalid_token') {
+        // Redirect user to login page perhaps?
+        console.log('SocketIO client user token has expired');
```
LucasBrazi06, line 41: Can't you test this on your local dev env?

jerome-benoit: That code path will have to stay to force a logout on the dashboard after being sure that everything works as expected everywhere. I can't test all browsers/OS combinations and I do not want to logout users because of bad support of socket.io code. The dashboard is still usable without proper socket.io support.

LucasBrazi06: > That code path will have to stay to force a logout on the dashboard after being sure that everything works as expected everywhere. I can't test all browsers/OS combinations and I do not want to logout users because of bad support of socket.io code. The dashboard is still usable without proper socket.io support.

i meant tested at least once on your side with one browser in your dev env?

jerome-benoit: > > That code path will have to stay to force a logout on the dashboard after being sure that everything works as expected everywhere. I can't test all browsers/OS combinations and I do not want to logout users because of bad support of socket.io code. The dashboard is still usable without proper socket.io support.
> 
> i meant tested at least once on your side with one browser in your dev env?

Sure. That's not because I have overlooked one time to do testing on a particular change that required a real charger not available at the time because of lab installation to validate it that you have to think that I don't do unit tests run and tests before PRs ;)

LucasBrazi06: > Sure. That's not because I have overlooked one time to do testing on a particular change that required a real charger not available at the time because of lab installation to validate it that you have to think that I don't do unit tests run and tests before PRs ;)
I'm just asking and expect yes/no and not a roman ;-).

PR #1791, opened 2020-07-03
Charging Station > OCPP Parameters > Regression: Put back two missing feature

WillyLellouch:
Added back two missing features for the ocpp parameters listing:
Get configuration from template (calls the last db version of the parameters)
Update parameters (Update db from charger then request the new version of the parameters)


src/app/pages/charging-stations/table-actions/table-request-ocpp-params-action.ts
```diff
@@ -0,0 +1,58 @@
...
+          spinnerService.hide();
+          if (response.status === OCPPGeneralResponse.ACCEPTED) {
+            messageService.showSuccessMessage(
+              translateService.instant('chargers.ocpp_params_update_from_template_success', { chargeBoxID: chargingStation.id }));
+            if (refresh) {
+              refresh().subscribe();
+            }
+          } else {
+            Utils.handleError(JSON.stringify(response),
+              messageService, 'chargers.ocpp_params_update_from_template_error');
```
OliveGerste, line 48: The message says: 'Error occurred while updating the OCPP parameters of the charging station' 
But you only request it from the station right?

src/app/pages/charging-stations/table-actions/table-request-ocpp-params-action.ts
```diff
@@ -0,0 +1,58 @@
...
+          }
+        }, (error: any) => {
+          spinnerService.hide();
+          Utils.handleHttpError(error, router, messageService,
+            centralServerService, 'chargers.ocpp_params_update_from_template_error');
```
OliveGerste, line 53: same as above.

src/app/pages/charging-stations/table-actions/table-request-ocpp-params-action.ts
WillyLellouch: changed messge to reflect actual process

src/app/pages/charging-stations/table-actions/table-request-ocpp-params-action.ts
WillyLellouch: see above

WillyLellouch:
> 
> 
> @WillyLellouch
> Looks okay to me for the front end site. (I would just change the two messages, which I commented).
> 
> One thing I noticed is, that there is no check from the back end site if the force update was accepted (correct me if I am wrong). So it will always show success even if it the update was rejected by the station.

I believe that happens only if there is no charger connected
If there is one, the ocpp request is denied and the server side sends out the appropriate error

jerome-benoit:
> > @WillyLellouch
> > Looks okay to me for the front end site. (I would just change the two messages, which I commented).
> > One thing I noticed is, that there is no check from the back end site if the force update was accepted (correct me if I am wrong). So it will always show success even if it the update was rejected by the station.
> 
> I believe that happens only if there is no charger connected
> If there is one, the ocpp request is denied and the server side sends out the appropriate error

If the charger is disconnected, the OCPP client will refuse to forward the request to it. Just make sure that the error propagation reach your code and that you handle it properly.

If a charger is connected, make sure you handle properly the OCPP message 'Rejected' or 'Accepted'.

PR #1908, opened 2020-08-07
Smart charging  add a new field amps per phase in edit charging station

OliveGerste:
I left the validation of the totalAmps, but I can also take it out if you want.


LucasBrazi06:
Good Job!

PR #1923, opened 2020-08-20
Badges management

mahdibenromdhane:


src/app/services/central-server-notification.service.ts
```diff
@@ -303,6 +313,14 @@ export class CentralServerNotificationService {
...
       this.subjectCar.next(changeNotification);
     });
+    // Monitor Tags
+    this.socketIOClient.socket.on(Entity.TAGS, (changeNotification: ChangeNotification) => {
+      this.subjectTags.next(changeNotification);
+    });
+    // Monitor Tag
+    this.socketIOClient.socket.on(Entity.TAG, (changeNotification: SingleChangeNotification) => {
+      this.subjectTag.next(changeNotification);
+    });
```
jerome-benoit, line 323: There's is non matching code in ev-server to send tags change notifications through socket.io.

LucasBrazi06: > There's is non matching code in ev-server to send tags change notifications through socket.io.

Ok Thanks!

PR #1926, opened 2020-08-24
Add a debug configuration section for browser console logging. 

jerome-benoit:
- Spell fixes;
- Comment out the dashboard link, not working at all;
- Optimise the configuration memory occupation by using a singleton like design pattern.

Code cleanups PR related to the work done to troubleshoot the connectivity issue on SCP (issue is closed so no linking is possible).


src/app/services/config.service.ts
```diff
@@ -17,66 +18,83 @@ import UserConfiguration from '../types/configuration/UserConfiguration';
...
-    this.load().catch(() => {});
+  constructor(private http?: HttpClient) {
+    this.load();
   }
 
-  public async load() {
-    this.config = await this.http.get<Configuration>('/assets/config.json').toPromise();
+  private getConfig(): Configuration {
+    if (!ConfigService.config) {
+      this.http.get<Configuration>('/assets/config.json').subscribe((configuration) => ConfigService.config = configuration);
```
LucasBrazi06, line 29: This is no longer a blocking call and you are returning 'null' if the file is not yet loaded

jerome-benoit: The rxjs subscribe method on observable will ensure that will never happen. 
It's the documented best practice on how to use angular http client to load local configuration files: https://angular.io/guide/http

LucasBrazi06: Last time I checked, subscribe() is asynchronous and is then a non blocking call.

jerome-benoit: subscribe() has the signature `subscribe(observer?: PartialObserver<T>): Subscription`, thus is synchronous and can be safely using here. 

Reference documentation: https://rxjs-dev.firebaseapp.com/api/index/class/Observable

LucasBrazi06: This code proves otherwise, have you tested it?
method load2() returns before the request is finished to be process so it's non blocking call.

image

jerome-benoit: It's tested and working. And it's the documented default way to handle local configuration files in Angular. 
I will retest it in a while by generating some load on the my local installation to remove any doubts. Let's wait that test results for the merge, Ok ?   

Your test case is testing the HTTP Angular client synchronicity, which is in fact asynchronous, but the variable affectation is done by the synchronous method subscribe() in the synchronous callback. To test to that the variable containing the configuration is properly affected synchronously, you have to go inside the callback. 
And you should remove async flag on load2().

LucasBrazi06: The issue is that the load should be done synchronously but it's done in the constructor (also in the former version with the Promise) and we cannot do sync calls in it!

In the end your call will work with the same side effect as the Promise call before because of the initial call from the constructor.

We still may have issues if the code requests a configuration param while the file is still loading (slow network).
You can give it a try by simulating a slow network on the browser.

jerome-benoit: > In the end your call will work with the same side effect as the Promise call before because of the initial call from the constructor.

I think I can get rid of the load() method in that class now, it just calls getConfig() ... 

I will simulate slow a network during the load test.

jerome-benoit: No configuration file load timing issue during the load test.

PR #1975, opened 2020-09-24
Badges management

mahdibenromdhane:


LucasBrazi06 (approved):
Just one issue remains: when URL below refreshed, the tab Tag is not automatically selected but Users is:
http://slf.localhost:45000/users#tag
Double check.

PR #2015, opened 2020-10-08
Default tag property on tag component

mahdibenromdhane:


LucasBrazi06 (approved):
I added the default tag column in the tag list
Put the Default checkbox to the end of the create/edit Tag pop-up

PR #2029, opened 2020-10-16
 Apply security update.

jerome-benoit:
And make whitesource happy.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -650,7 +650,8 @@
       },
       "dependencies": {
         "ajv": {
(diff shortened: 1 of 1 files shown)


LucasBrazi06:
> And make whitesource happy.

We have not yet run a scan on the ev-dashboard :-)

jerome-benoit:
> > And make whitesource happy.
> 
> We have not yet run a scan on the ev-dashboard :-)

Yep it will be happier if we proactively run `npm audit fix` before ;)

LucasBrazi06:
> Yep it will be happier if we proactively run `npm audit fix`;)

Hahaha, yes! Kill it in the egg! 
Thanks!

PR #2046, opened 2020-10-27
Auth Rest API Implementation

Ninjeneer:


src/app/services/central-server.service.ts
```diff
@@ -2366,7 +2366,7 @@ export class CentralServerService {
     // Set the tenant
     params['Tenant'] = this.windowService.getSubdomain();
     // Execute the REST service
-    return this.httpClient.get<ActionResponse>(`${this.centralRestServerServiceAuthURL}/${ServerAction.VERIFY_EMAIL}`,
+    return this.httpClient.get<ActionResponse>(`${this.restServerAuthURL}/${ServerAction.VERIFY_EMAIL}`,
```
jerome-benoit, line 2369: That endpoint does not exist on the new REST API. 

And the ServerAction does not follow the naming scheme defined in the issue relative to the new API definition. 

Have you tested your changes? because I doubt it works.

src/app/services/central-server.service.ts
```diff
@@ -3004,7 +3004,7 @@ export class CentralServerService {
       // Set REST base URL
       this.centralServerNotificationService.setcentralRestServerServiceURL(this.centralRestServerServiceBaseURL);
       // Auth API
-      this.centralRestServerServiceAuthURL = this.centralRestServerServiceBaseURL + '/client/auth';
+      this.restServerAuthURL = this.centralRestServerServiceBaseURL + '/v1/auth';
```
jerome-benoit, line 3007: Idem.

jerome-benoit (changes requested):

jerome-benoit (commented):
For consistency, you can maybe rename some auth related routes similarly to the one used in the new REST API.

LucasBrazi06 (changes requested):
Same remark as in backend:
Use full path

jerome-benoit (approved):

PR #2079, opened 2020-11-10
Export CDR Json document in session list

WillyLellouch:
Front-end support for CDR document export (OCPI Data Export)

Test: 
1. Row action "Export Ocpi Data" should only appear in "More actions" on transactions with OcpiData field such as transactions from FR*007 users
2. When exporting JSON file, compare with database : content of file should be identical to the content of OcpiData Field in corresponding transaction.


LucasBrazi06:
@WillyLellouch Resolve the conflicts

PR #2144, opened 2020-12-09
Notification: Fix SMTP error handling to avoid useless sending tries

jerome-benoit:


LucasBrazi06:
@jerome-benoit 
After the front-end changes I decided that the impact of a simple variable renaming is too vast.
Could you please revert this change?
Thanks.

jerome-benoit:
> @jerome-benoit
> After the front-end changes I decided that the impact of a simple variable renaming is too vast.
> Could you please revert this change?
> Thanks.

I've by mistake stacked PRs on ev-dashboard, so if you merge https://github.com/LucasBrazi06/ev-dashboard/pull/2117, that one become more readable. So will become https://github.com/LucasBrazi06/ev-dashboard/pull/2145 after the merge of that one. 

Variables, classes, methods, etc. rename is ev-dashboard are harmless. So they are on ev-server. We do that all the times. The rename done here were done automatically by code and just works. 

The only rename that have an impact is the one done on the User type because it's a REST API change. And reverting it just change 3 lines in the current PR and make the code less readable. 
Reverting it on ev-server change also a few lines and for safety, I can add both fields to the REST API endpoints. 

Since there's no hurry on adding SMTP error handling to trigger correctly SMTP backup use, that PRs can be merged only when a full test plan of that feature can be scheduled with Willy first.

LucasBrazi06:
> I've by mistake stacked PRs on ev-dashboard, so if you merge #2117, that one become more readable. So will become #2145 after the merge of that one.
> 
> Variables, classes, methods, etc. rename is ev-dashboard are harmless. So they are on ev-server. We do that all the times. The rename done here were done automatically by code and just works.
> 
> The only rename that have an impact is the one done on the User type because it's a REST API change. And reverting it just change 3 lines in the current PR and make the code less readable.
> Reverting it on ev-server change also a few lines and for safety, I can add both fields to the REST API endpoints.
> 
> Since there's no hurry on adding SMTP error handling to trigger correctly SMTP backup use, that PRs can be merged only when a full test plan of that feature can be scheduled with Willy first.

Ok then you take full responsibility of this choice and will report directly to the boss in case of failure.

.pipeline/Acceptance.groovy
```diff
@@ -0,0 +1,21 @@
...
+    userRemoteConfigs: [
+        [credentialsId: 'GitHub-emobility-ci-build', url: "https://github.wdf.sap.corp/EVSE/ev-scp-sap-it.git"]
+    ]
+  ])
+  sh "ls"
+  sh "ls deployment"
+  //params.originalStage()
+
+}
+return this
```
LucasBrazi06, line 21: What has it to do with the current issue!?

LucasBrazi06 (approved):

PR #2159, opened 2020-12-22
OICP Settings implementation

Jakob8:
New Roaming component for Hubject (OICP Protocol)
Enable Roaming in Tenant Settings and add Roaming Settings
Register Hubject Endpoint
Start background jobs


LucasBrazi06 (changes requested):
I think that many components like formatters could have been put in common with OCPI.

PR #2161, opened 2020-12-23
Smart charging   additional settings

OliveGerste:


src/app/types/Setting.ts
```diff
@@ -121,6 +121,9 @@ export interface SapSmartChargingSetting {
   optimizerUrl: string;
   user: string;
   password: string;
+  stickyLimitation: boolean;
+  limitBufferDC: number;
```
LucasBrazi06, line 125: For the next PR, it would be better to name the unit I think: limitBufferDCPercent

src/app/types/Setting.ts
```diff
@@ -121,6 +121,9 @@ export interface SapSmartChargingSetting {
...
+  limitBufferAC: number;
```
LucasBrazi06, line 126: Do we need a different buffer for AC also?

LucasBrazi06 (approved):

PR #2165, opened 2020-12-26
Add Active filter + Multiselect delete tags

mahdibenromdhane:


src/app/pages/users/tags/tags-list-table-data-source.ts
```diff
@@ -135,11 +142,35 @@ export class TagsListTableDataSource extends TableDataSource<Tag> {
     });
   }
 
+
+  public toggleRowSelection(row: Tag, checked: boolean) {
+    super.toggleRowSelection(row, checked);
+    this.deleteManyAction.disabled = this.selectedRows === 0;
```
LucasBrazi06, line 148: Enabling/Disabling a toolbar button according an item selection in the list should be done in a generic way by adding a new property in TableActionDef like 'linkedToListSelection: boolean' and use this prop in TableDataSource class to enable/disable automatically all these buttons.
You can also apply this change to all views in another issue.
Thanks!

src/assets/i18n/de.json
```diff
@@ -1273,7 +1273,12 @@
     "tag_id_already_used": "Der Badge '{{tagID}}' existiert bereits",
     "redirect": "Zu den Badges navigieren",
     "title": "Badge",
-    "default_tag": "Default badge"
+    "default_tag": "Default badge",
+    "delete_tags_title": "Delete Badge(s)",
+    "delete_tags_confirm": "Do you really want to delete '{{quantity}}' badge(s)?",
+    "delete_tags_success": "'{{inSuccess}}' badges have been deleted successfully",
+    "delete_tags_partial": "'{{inSuccess}}' badges  have been deleted successfully and '{{inError}}' badges have encountered an error, check the logs",
```
LucasBrazi06, line 1280: Extra space to remove.

src/assets/i18n/en.json
```diff
@@ -1273,7 +1273,12 @@
     "tag_id_already_used": "The badge '{{tagID}}' already exists",
     "redirect": "Navigate to Tags",
     "title": "Badge",
-    "default_tag": "Default badge"
+    "default_tag": "Default badge",
+    "delete_tags_title": "Delete Badge(s)",
+    "delete_tags_confirm": "Do you really want to delete '{{quantity}}' badge(s)?",
+    "delete_tags_success": "'{{inSuccess}}' badges have been deleted successfully",
+    "delete_tags_partial": "'{{inSuccess}}' badges  have been deleted successfully and '{{inError}}' badges have encountered an error, check the logs",
```
LucasBrazi06, line 1280: Extra space to remove.

src/assets/i18n/es.json
```diff
@@ -1273,7 +1273,12 @@
     "tag_id_already_used": "The badge '{{tagID}}' already exists",
     "redirect": "Navigate to Tags",
     "title": "Badge",
-    "default_tag": "Default badge"
+    "default_tag": "Default badge",
+    "delete_tags_title": "Delete Badge(s)",
+    "delete_tags_confirm": "Do you really want to delete '{{quantity}}' badge(s)?",
+    "delete_tags_success": "'{{inSuccess}}' badges have been deleted successfully",
+    "delete_tags_partial": "'{{inSuccess}}' badges  have been deleted successfully and '{{inError}}' badges have encountered an error, check the logs",
```
LucasBrazi06, line 1280: Extra space to remove.

src/assets/i18n/pt.json
```diff
@@ -1273,7 +1273,12 @@
     "tag_id_already_used": "O badge '{{tagID}}' já existe",
     "redirect": "Navegue até as tags",
     "title": "Badge",
-    "default_tag": "Badge padrão"
+    "default_tag": "Badge padrão",
+    "delete_tags_title": "Delete Badge(s)",
+    "delete_tags_confirm": "Do you really want to delete '{{quantity}}' badge(s)?",
+    "delete_tags_success": "'{{inSuccess}}' badges have been deleted successfully",
+    "delete_tags_partial": "'{{inSuccess}}' badges  have been deleted successfully and '{{inError}}' badges have encountered an error, check the logs",
```
LucasBrazi06, line 1280: Extra space to remove.

LucasBrazi06 (changes requested):
Check my comments

src/assets/i18n/de.json
jerome-benoit: @mahdibenromdhane: if you install the recommended plugins in vscode and ensure they are all enabled, you not have that kind of formatting issue.

mahdibenromdhane: > @mahdibenromdhane: if you install the recommended plugins in vscode and ensure they are all enabled, you will not have that kind of formatting issue.

I thought I have them, I will double check. Thanks.

src/app/pages/users/tags/tags-list-table-data-source.ts
mahdibenromdhane: Done for this case, I will create an issue for the other cases.

src/app/shared/table/table.component.html
```diff
@@ -9,7 +9,7 @@
           <!-- Action Button -->
           <button mat-raised-button *ngIf="actionDef.type === 'button' && actionDef.id !== 'reset_filters'"
             (click)="actionTriggered(actionDef)" [color]="(actionDef.color ? actionDef.color : '')"
-            [disabled]="actionDef.disabled" [title]="(actionDef.name ? '' : actionDef.tooltip | translate)">
+    [disabled]="actionDef.disabled|| (dataSource.tableDef.rowSelection && actionDef.linkedToListSelection && dataSource.selectedRows === 0)" [title]="(actionDef.name ? '' : actionDef.tooltip | translate)">
```
LucasBrazi06, line 12: I don't think all the buttons are related to row selection (eg. export)
So this is why I think we need a new type, right?

LucasBrazi06 (changes requested):

src/app/pages/users/tags/tags-list-table-data-source.ts
LucasBrazi06: Not gut ;-)
Check my comments below

src/app/shared/table/table.component.html
mahdibenromdhane: > I don't think all the buttons are related to row selection (eg. export)
> So this is why I think we need a new property as describe in the other comment, right?
> 
> > Enabling/Disabling a toolbar button according an item selection in the list should be done in a generic way by adding a new property in TableActionDef like 'linkedToListSelection: boolean' and use this prop in TableDataSource class to enable/disable automatically all these buttons.

I'v added the property linkedToListSelection as requested so we can distinguish them from those whom are not related to the row selection.

PR #2182, opened 2021-01-08
Add configuration to specify the landscape

jerome-benoit:
And use it to disable firmware upgrade tab on production.

Update: 

+ Move helpers to an injectable service class.

Signed-off-by: Jérôme Benoit <jerome.benoit@sap.com>


src/app/utils/Utils.ts
```diff
@@ -860,4 +861,19 @@ export class Utils {
     document.execCommand('copy');
     document.body.removeChild(element);
   }
+
+  public static isDevLandscape(): boolean {
+    const configService: ConfigService = new ConfigService();
```
LucasBrazi06, line 866: Should not be instantiated like this.
Inject the config service where needed and test the Landscape here or create an Utils Service that provides these helper methods which will in the end be used the same way to hide the tab.

LucasBrazi06 (changes requested):

PR #2200, opened 2021-01-14
02. Crypto - Technical settings

stefania-santimbrean:
Connected to ev-server PR #2038 

Did the PR for partial review (since it's my first PR on the dashboard).


LucasBrazi06 (approved):
Naming convention not followed
Tab space size is 4 instead of 2
Disable controls from component and not in HTML
Other i18n language files not maintained

PR #2209, opened 2021-01-19
Move charging station

Ninjeneer:
Still need a translation for ES / DE / PT


jerome-benoit:
The feature to have a UI to easily move charging stations from one site area to another have been requested by at least 3 customers: they do the integration and configuration at their desk then ship the charging station to the destination site area, install and validate it, create the target site area and then move the station to the target site area.

LucasBrazi06:
> The feature to have a UI to easily move charging stations from one site area to another have been requested by at least 3 customers: they do the integration and configuration at their desk then ship the charging station to the destination site area, install and validate it, create the target site area and then move the station to the target site area.

Edit Charging Station > change the Site Area > Save ?

jerome-benoit:
> > It's a counter intuitive user experience design, the action is not explicit.
> 
> Okay, I think we have at least a work around that we can explain to the customer.

We (Kinza, Gerald and I) will redirect the support flow generated by counter intuitive user experience design directly to you in the meantime :)

LucasBrazi06:
> We (Kinza, Gerald and I) will redirect the support flow generated by counter intuitive user experience design directly to you in the meantime :)

Okay, I will discuss with them then :-)

PR #2210, opened 2021-01-19
POC filters dependencies. 

mahdibenromdhane:
- Passing Issuer filter to Tag filter in users list
- Passing Site filter to SiteArea filter in charging stations list


src/app/pages/charging-stations/list/charging-stations-list-table-data-source.ts
```diff
@@ -325,11 +325,12 @@ export class ChargingStationsListTableDataSource extends TableDataSource<Chargin
...
   public buildTableFiltersDef(): TableFilterDef[] {
     if (this.isOrganizationComponentActive) {
+      const siteFilter = new SiteTableFilter().getFilterDef();
       return [
         // new ChargingStationTableFilter().getFilterDef(),
         new IssuerFilter().getFilterDef(),
-        new SiteTableFilter().getFilterDef(),
-        new SiteAreaTableFilter().getFilterDef()
+        siteFilter,
+        new SiteAreaTableFilter(null, [siteFilter]).getFilterDef(),
```
LucasBrazi06, line 333: sideIDs should be removed from the SiteAreaTableFilter and replaced by the filter deps.

src/app/shared/table/filters/site-area-table-filter.ts
```diff
@@ -3,7 +3,7 @@ import { FilterType, TableFilterDef } from '../../../types/Table';
 import { TableFilter } from './table-filter';
 
 export class SiteAreaTableFilter extends TableFilter {
-  constructor(siteIDs?: ReadonlyArray<string>) {
+  constructor(siteIDs?: ReadonlyArray<string>, dependentFilters?: TableFilterDef[]) {
```
LucasBrazi06, line 6: siteIDs param to remove

LucasBrazi06 (changes requested):
The concept is validated!
Check my comments for the rest.

src/app/pages/charging-stations/list/charging-stations-list-table-data-source.ts
mahdibenromdhane: Done

src/app/shared/table/filters/site-area-table-filter.ts
mahdibenromdhane: Done

LucasBrazi06 (approved):
Gut!
PoC is validated.
You can generalize the feature to all other filters in all the app.
We have to think also about this use case: when the Site is changed, Site Area filter should be cleared :-)
I'll create a new issue for this.

mahdibenromdhane:
> Gut!
> PoC is validated.
> You can generalize the feature to all other filters in all the app.
> We have to think also about this use case: when the Site is changed, Site Area filter should be cleared :-)
> I'll create a new issue for this.

Great :)

Issue #2045, opened 2020-10-27
Charging Station: Add Unlock Connector OCPP feature in the Dashboard


LucasBrazi06:

jerome-benoit:
+ The unlock action should not appear for domestic plug;
+ Unlocking trigger an harmless and misleading error message in the log about unlocking failure.

PR #2215, opened 2021-01-20
03. Crypto - Change crypto setting in UI

stefania-santimbrean:
Connected to ev-server [PR #2045](https://github.com/LucasBrazi06/ev-server/pull/2045)


src/app/pages/settings-technical/crypto/settings-crypto.component.ts
```diff
@@ -64,51 +64,55 @@ export class SettingsCryptoComponent implements OnInit {
...
-        this.spinnerService.hide();
-        switch (error.status) {
-          case HTTPError.OBJECT_DOES_NOT_EXIST_ERROR:
-            this.messageService.showErrorMessage('technical_settings.crypto.setting_do_not_exist');
-            break;
-          default:
-            Utils.handleHttpError(
-              error,
-              this.router,
+    if (this.cryptoSettings.crypto.migrationToBeDone) {
```
LucasBrazi06, line 67: No, the request has to be sent to the backend and the backend will return an HTTP error code and you will display the right error message.
Check HTTPError type.

src/assets/i18n/en.json
```diff
@@ -1565,7 +1565,8 @@
       },
       "setting_do_not_exist": "Crypto Technical Settings not found",
       "update_success": "Crypto Technical Settings have been updated successfully",
-      "update_error": "Error occurred while updating the Crypto Technical Settings, check the logs"
+      "update_error": "Error occurred while updating the Crypto Technical Settings, check the logs",
+      "migration_in_progress_warning": "Crypto Settings cannot be changed while migration is in progress."
```
LucasBrazi06, line 1569: Have you changed also all other i18n !?
Golden rule #4

LucasBrazi06 (changes requested):

LucasBrazi06 (approved):

LucasBrazi06:
@jerome-benoit It seems that the UI does not fully enforce values below!!! I entered a long key and made the migration failed and blocked!!!

I need your advice here.
Should we make the block size/cypher/operation mode read-only?
Also the key may also change according the fields below.
Let me know so I'll add the missing assertions in the UI and backend.
Thanks in advance!

LucasBrazi06 (commented):
I would need more check on UI element like check of the key length, algorithm...

jerome-benoit:
> @jerome-benoit <https://github.com/jerome-benoit> It seems that the UI does not fully enforce values below!!! I entered a long key and made the migration failed and blocked!!!
> 
> I need your advice here.
> Should we make the block size/cypher/operation mode read-only?
> Also the key may also change according the fields below.
> Let me know so I'll add the missing assertions in the UI and backend.
> Thanks in advance!
> 

It should have been dropdown manus depending on the symmetric cypher algo.
The secure settings are aes 256 gcm and it’s currently the default. They can stay read only as long as at key change, the default are shown instead of the current ones.

--
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

LucasBrazi06:
> It should have been dropdown manus depending on the symmetric cypher algo. The secure settings are aes 256 gcm and it’s currently the default. They can stay read only as long as at key change, the default are shown instead of the current ones.

Ok, let the user changes only the key then for the time being.
We can propose later on a limited list of algo that we allow in a new drop-down.
Btw, default is aes-256-ctr (not aes-256-gcm):

image

LucasBrazi06:
@jerome-benoit also you can limit the list of supported algos in Constants class:

image

jerome-benoit:
> Btw, default is aes-256-ctr (not aes-256-gcm):
> 
>  <https://user-images.githubusercontent.com/25266945/109421259-db0da980-79d6-11eb-99bc-dec717a47f93.png>
> 

That’s not the secure default I’ve request during the development, it shall be aes-256-gcm.

--
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

LucasBrazi06:
> That’s not the secure default I’ve request during the development, it shall be aes-256-gcm.

She had done gcm but this was not working with current configuration, not tested in new tenant.
So I think we need another migration task for this one.

PR #2256, opened 2021-02-18
renaming: pushLocation -> pushEVSEStatuses

Jakob8:


Jakob8:
pushLocations is wrong name because only the EVSE statuses (no locations) are pushed.
renaming: pushLocation -> pushEVSEStatuses

LucasBrazi06:
PR not attached to the issue.
Just did it.

PR #2269, opened 2021-02-24
added markAs Dirty/Pristine after update site only

cabch:
added error handling for site admin (cannot create token without site)
modified actions/more actions displaying
added BAD_REQUEST error
updated i18n for bad request error on token for site admin

### DIFF
--- a/src/app/pages/charging-stations/registration-tokens/registration-token/charging-stations-registration-token.component.ts
+++ b/src/app/pages/charging-stations/registration-tokens/registration-token/charging-stations-registration-token.component.ts
@@ -74,6 +74,7 @@ export class ChargingStationsRegistrationTokenComponent implements OnInit {
     if (this.currentTokenID) {
       this.spinnerService.show();
       this.centralServerService.getRegistrationToken(this.currentTokenID).subscribe((registrationToken) => {
+        this.formGroup.markAsPristine();
         this.spinnerService.hide();
         this.currentToken = registrationToken;
         // Init form
@@ -128,7 +129,13 @@ export class ChargingStationsRegistrationTokenComponent implements OnInit {
       }
     }, (error) => {
       this.spinnerService.hide();
-      Utils.handleHttpError(error, this.router, this.messageService, this.centralServerService, 'tenants.create_error');
+      switch (error.status) {
+        case HTTPError.BAD_REQUEST:
+          this.messageService.showErrorMessage('chargers.connections.registration_token_site_admin_creation_error');
+          break;
+        default:
+          Utils.handleHttpError(error, this.router, this.messageService, this.centralServerService, 'chargers.connections.registration_token_creation_error');
+      }
     });
   }
 
@@ -177,6 +184,7 @@ export class ChargingStationsRegistrationTokenComponent implements OnInit {
         const siteArea = (result[0].objectRef) as SiteArea;
         this.siteArea.setValue(`${(siteArea.site ? siteArea.site.name + ' - ' : '')}${siteArea.name}`);
         this.siteAreaID.setValue(siteArea.id);
+        this.formGroup.markAsDirty();
       }
     });
   }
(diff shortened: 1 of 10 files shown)


src/app/pages/charging-stations/registration-tokens/registration-token/charging-stations-registration-token.component.ts
```diff
@@ -128,7 +129,13 @@ export class ChargingStationsRegistrationTokenComponent implements OnInit {
       }
     }, (error) => {
       this.spinnerService.hide();
-      Utils.handleHttpError(error, this.router, this.messageService, this.centralServerService, 'tenants.create_error');
+      switch (error.status) {
+        case HTTPError.BAD_REQUEST:
```
LucasBrazi06, line 133: It's not BAD_REQUEST I think, it's FORBIDDEN ;-)

src/app/types/HTTPError.ts
```diff
@@ -60,6 +60,7 @@ export enum HTTPError {
 
   CANNOT_ACQUIRE_LOCK = 510,
   INVALID_FILE_FORMAT = 583,
+  BAD_REQUEST = 400,
```
LucasBrazi06, line 63: No need this BAD_REQUEST.
More over don't confuse HttpError with HttpAuthError.

src/app/utils/Utils.ts
```diff
@@ -752,8 +752,6 @@ export class Utils {
         messageService.showErrorMessage('general.not_authorized');
         break;
       case StatusCodes.BAD_REQUEST:
-        messageService.showErrorMessage('general.invalid_content');
```
LucasBrazi06, line 755: Use FORBIDDEN

LucasBrazi06 (changes requested):

src/app/pages/charging-stations/registration-tokens/registration-token/charging-stations-registration-token.component.ts
cabch: modified to forbidden

src/app/types/HTTPError.ts
cabch: deleted and modified error handling to forbidden

src/app/utils/Utils.ts
cabch: modified

src/app/pages/charging-stations/registration-tokens/registration-token/charging-stations-registration-token.component.ts
```diff
@@ -128,7 +129,13 @@ export class ChargingStationsRegistrationTokenComponent implements OnInit {
       }
     }, (error) => {
       this.spinnerService.hide();
-      Utils.handleHttpError(error, this.router, this.messageService, this.centralServerService, 'tenants.create_error');
+      switch (error.status) {
+        case HTTPAuthError.FORBIDDEN:
```
LucasBrazi06: Normally if it's forbidden you should let the Utils.handleHttpError() display the message.

LucasBrazi06 (approved):

src/app/pages/charging-stations/registration-tokens/registration-token/charging-stations-registration-token.component.ts
cabch: what if we want to display a personnalized message letting the user know it misses the site ? because otherwise he will see something like "you cannot do this" without knowing this is only a matter of missing param ?

LucasBrazi06: > what if we want to display a personnalized message letting the user know it misses the site ? because otherwise he will see something like "you cannot do this" without knowing this is only a matter of missing param ?

You should not have a personalised message as the action is forbidden and you should not even see the action itself ;-)

PR #2279, opened 2021-03-01
Checkbox for notifications activate user account manually

cabch:
- added checkbox for admin notification (true by default = receive the admin email)

- modified if/else for route users/:id as it was entering the else statement even when we weren't in the /:id route and it was making a call to getUser (with undefined userId) unnecessarily 
- deleted "clear search" as it's not needed anymore


LucasBrazi06 (approved):
Perfect!

PR #2305, opened 2021-03-16
billing - settings - change the order of stripe fields to avoid unwan…

ClaudeROSSI:
billing - settings - change the order of stripe fields to avoid unwanted chrome autocomplete features

### DIFF
--- a/src/app/pages/settings-integration/billing/settings-billing.component.html
+++ b/src/app/pages/settings-integration/billing/settings-billing.component.html
@@ -10,7 +10,7 @@
     {{'settings.activation_contact_msg' | translate}}
   </div>
 </div>
-<form *ngIf="isActive" class="form" [formGroup]="formGroup">
+<form *ngIf="isActive" class="form" [formGroup]="formGroup" autocomplete="off">
   <div class="mat-toolbar-row left-actions">
     <button mat-raised-button color="warn" (click)="save(formGroup.value)"
       [disabled]="!formGroup.valid || !formGroup.dirty">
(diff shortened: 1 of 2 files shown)


ClaudeROSSI:
Does not work as expected ...

@cabch, @mahdibenromdhane : I need your advise

ClaudeROSSI:
The change I have done is simply to make sure the input fields do not get populated with the user name and regular user password when opening the Billing settings. This to prevent a confusing behavior.

However, I couldn't find any option to prevent chrome from considering the STRIPE settings as providing user/password information.
The side-effect (at least on my machine) is that chrome keep asking the user whether he/she wants to save the new user/password.
This is an issue because the public_key and the secret_key is then somehow persisted client side.

LucasBrazi06:
> The change I have done is simply to make sure the input fields do not get populated with the user name and regular user password when opening the Billing settings. This to prevent a confusing behavior.
> 
> However, I couldn't find any option to prevent chrome from considering the STRIPE settings as providing user/password information.
> The side-effect (at least on my machine) is that chrome keep asking the user whether he/she wants to save the new user/password.
> This is an issue because the public_key and the secret_key is then somehow persisted client side.

You need to show me ;-)

ClaudeROSSI:
> > The change I have done is simply to make sure the input fields do not get populated with the user name and regular user password when opening the Billing settings. This to prevent a confusing behavior.
> > However, I couldn't find any option to prevent chrome from considering the STRIPE settings as providing user/password information.
> > The side-effect (at least on my machine) is that chrome keep asking the user whether he/she wants to save the new user/password.
> > This is an issue because the public_key and the secret_key is then somehow persisted client side.
> 
> You need to show me ;-)

Actually the change do not work. For some reasons Chrome ignores the settings!
I get a popup to save the secret_key as my new password each time i change one of the property of the billings settings form!
This is crazy!

ClaudeROSSI:
We need to find another solution ... putting it back to in progress

PR #2348, opened 2021-04-08
Billing - UI - consume new REST endpoints

ClaudeROSSI:
Operations targeting the payment methods are now targeting the new REST endpoints


package.json
```diff
@@ -110,6 +110,7 @@
     "socket.io-client": "^3.1.3",
     "sweetalert2": "^10.15.6",
     "tslib": "^2.1.0",
+    "url-pattern": "^1.0.3",
```
jerome-benoit, line 113: I would prefer that you use a widely used URL parser like parseurl. I would even prefer that you rely only on express URI parameters extraction but that seems difficult here. 

You can use https://www.npmtrends.com/url-pattern-vs-parseurl-vs-parseuri-vs-url-parse to compare npm package usage.

ClaudeROSSI: thanks for the hints ... i'll have a look. we also need to escape url fragments. Right now this is not done.

ClaudeROSSI: > I would prefer that you use a widely used URL parser like parseurl.
> 
> You can use https://www.npmtrends.com/url-pattern-vs-parseurl-vs-parseuri-vs-url-parse to compare npm package usage.

Well ... I do not need an URL parser. I simply need an URL builder.
The one matching my needs is: https://www.npmjs.com/package/safe-url-assembler
Are you aware of an alternative? How to generate URLs matching endpoints defined using dynamic express routes such as: /things/:id/:name

jerome-benoit: No but go for that one then.

src/app/services/central-server.service.ts
```diff
@@ -3274,4 +3287,15 @@ export class CentralServerService {
     }
     return of(null);
   }
+
+  private buildRestEndpointUrl(urlPatternAsString: ServerRoute, params: {
+    // Just a flat list of key/value pairs!
+    [name: string]: string | number | null;
+  }) {
+    const url = SafeUrlAssembler(this.restServerSecuredURL)
+      .template('/'+urlPatternAsString)
```
jerome-benoit, line 3296: nitpick: space before and after `+`.

LucasBrazi06 (approved):
Check Jerome's comment.
Mergeable anyway :-)

PR #2364, opened 2021-04-16
Static limit - Renamed slider changed event 

OliveGerste:


jerome-benoit (approved):
I've introduced the regression ... Thanks for the fix.

LucasBrazi06:
> I've introduced the regression ... Thanks for the fix.

Thanks Tobias!
No problem, the most important thing is our reactivity when this happen ;-).

PR #2369, opened 2021-04-19
Authorizations > Front-end: Enhancement of the current concept (Sites)

Fonsu:


src/app/pages/organization/sites/site/site-dialog.component.ts
```diff
@@ -1,21 +1,23 @@
...
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
-import { DialogData } from 'types/Authorization';
+import { SiteDialogData } from 'types/Authorization';
 
 import { Utils } from '../../../../utils/Utils';
 import { SiteComponent } from './site.component';
 
 @Component({
-  template: '<app-site #appRef [currentSiteID]="siteID" [inDialog]="true" [dialogRef]="dialogRef"></app-site>',
+  template: '<app-site #appRef [currentSiteID]="siteID" [canUpdateSite]="canUpdateSite" [inDialog]="true" [dialogRef]="dialogRef"></app-site>',
```
LucasBrazi06, line 9: Should we pass the site so if we enrich again the object, we will have new props for free without touching template.

src/app/shared/table/actions/sites/table-assign-users-to-site-action.ts
```diff
@@ -1,13 +1,15 @@
...
 
-import { Site, SiteButtonAction } from '../../../../types/Site';
+import { SiteButtonAction } from '../../../../types/Site';
 import { TableActionDef } from '../../../../types/Table';
 import { TableAssignAction } from '../table-assign-action';
 
 export interface TableAssignUsersToSiteActionDef extends TableActionDef {
-  action: (siteUsersDialogComponent: ComponentType<unknown>, site: Site, dialog: MatDialog, refresh?: () => Observable<void>) => void;
+  action: (siteUsersDialogComponent: ComponentType<unknown>, dialog: MatDialog,
+    data: SiteDialogData, refresh?: () => Observable<void>) => void;
```
LucasBrazi06, line 12: site: Site?

src/app/types/Authorization.ts
```diff
@@ -158,6 +158,15 @@ export interface AuthorizationActions {
...
+export interface SiteAuthorizationActions extends AuthorizationActions {
+  canAssignUsers?: boolean;
+  canUnassignUsers?: boolean;
+}
+
 export interface DialogData extends AuthorizationActions {
   id?: string | number;
 }
+
+export interface SiteDialogData extends DialogData, SiteAuthorizationActions {
```
LucasBrazi06, line 170: Maybe not needed

src/app/types/Table.ts
```diff
@@ -27,6 +27,9 @@ export interface Data extends AuthorizationActions {
   isExpanded?: boolean;
 }
 
+export interface SiteData extends Data, SiteAuthorizationActions {
```
LucasBrazi06, line 30: We can keep only data as mother class of Site

src/app/types/Site.ts
```diff
@@ -24,7 +24,7 @@ export interface Site extends Data, CreatedUpdatedProps {
   public?: boolean;
 }
 
-export interface SiteUser extends Data {
+export interface SiteUser extends SiteData {
```
LucasBrazi06, line 27: We can keep data here I think. We can down/upcast the type in the code I think.

src/app/types/Site.ts
```diff
@@ -2,10 +2,10 @@ import { Address } from './Address';
 import { Company } from './Company';
 import CreatedUpdatedProps from './CreatedUpdatedProps';
 import { SiteArea } from './SiteArea';
-import { Data } from './Table';
+import { SiteData } from './Table';
 import { User } from './User';
 
-export interface Site extends Data, CreatedUpdatedProps {
+export interface Site extends SiteData, CreatedUpdatedProps {
```
LucasBrazi06, line 8: We can keep data here I think. We can down/upcast the type in the code I think.

LucasBrazi06 (changes requested):
I think we can pass the whole Site object to both pop-ups and if tomorrow we add new props, we'll benefit from getting them right from the object without changing anything in the dialog data / template.
Let me know if this is clear.
If not I can try to make a PR on top of yours to check if it's feasible.
If yes, you can have a look and merge it to yours and then I'll merge both the back and the front PRs together.

LucasBrazi06:
@Fonsu What about another solution to have a pop-up mode property with enums read, update and create to display the Site or Company pop-ups.
And for the other kind of pop-up like Assign Users we can pass the whole Site object where we can directly read values.
What do you think?

Fonsu:
Thank you for your review 🙂 

> I think we can pass the whole Site object to both pop-ups and if tomorrow we add new props, we'll benefit from getting them right from the object without changing anything in the dialog data / template.

That makes sense, that way we don't strip the `site.id` from `site` and pass it to the popup only to fetch the whole site from the backend again. Done in my last two commits. 

> @Fonsu What about another solution to have a pop-up mode property with enums read, update and create to display the Site or Company pop-ups.
> And for the other kind of pop-up like Assign Users we can pass the whole Site object where we can directly read values.
> What do you think?

Maybe 'enriching' `site` in `sites-list-table-data-source.ts:220` so that there is also a `mode` that can be checked on `site.component.html` would be useful. I am not entirely sure if that is what you mean, though. With this approach there would be no need to 'force' `canUpdate` to `false` in the view action, which would definitely improve readability.

LucasBrazi06:
> That makes sense, that way we don't strip the `site.id` from `site` and pass it to the popup only to fetch the whole site from the backend again. Done in my last two commits.

Thanks!

> Maybe 'enriching' `site` in `sites-list-table-data-source.ts:220` so that there is also a `mode` that can be checked on `site.component.html` would be useful. I am not entirely sure if that is what you mean, though. With this approach there would be no need to 'force' `canUpdate` to `false` in the view action, which would definitely improve readability.

You can check this when the action is triggered and set the PopupMode when calling it.
If it's ok for you, you can go ahead.

PR #2374, opened 2021-04-21
Create company site proposal

LucasBrazi06:
- Passing the Site's object to the Edit/Create pop-up is not a good idea in the end because not all the Site props are loaded in the list (even worse for other entities like Charging Station)
- I added a dialogMode propertie which will drive the dialog's state which is given by the list and the pop-up will trust it to enable/disable the formGroup
- I removed the 'inDialog' props as we have 'dialogRef' to know if we are in a dialog
- I added readOnly flag in the controller's pop-up to display/hide components in the HTML template
- I removed the check of the issuer flag in the component after reading the Site (already driven by the canUpdate flag in the list which will open the pop-up in read-only)

- Double check this PR and approve it if you are ok or make proposals.
- Test the full app (I did it but better do it twice)
- If okay, I'll merge then both PRs + front-end
- And you can report the same changes in Site Area in a new branch by doing the same thing than we did in Site

Thanks!

### DIFF
--- a/src/app/pages/assets/asset/asset.dialog.component.ts
+++ b/src/app/pages/assets/asset/asset.dialog.component.ts
@@ -1,6 +1,7 @@
 import { AfterViewInit, Component, Inject, ViewChild } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
-import { DialogData } from 'types/Authorization';
+import { Asset } from 'types/Asset';
+import { DialogParams } from 'types/Authorization';
 
 import { Utils } from '../../../utils/Utils';
 import { AssetComponent } from './asset.component';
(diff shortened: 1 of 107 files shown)


Fonsu (commented):
> * I removed the check of the issuer flag in the component after reading the Site (already driven by the canUpdate flag in the list which will open the pop-up in read-only)
> *

Thank you @LucasBrazi06 , this is a much cleaner and simpler approach! It keeps the component (i.e. Site) abstracted from specific permissions by just using the generic `readOnly` flag. I've been testing and didn't find issues so far. 🙂 

A couple minor considerations though:
* If `canUpdate` already considers `issuer`, isn't `sites-list-table-data-source.ts:189` redundant?
* Should we take into account `moreActions` such as `exportOCPPParamsAction` or `siteGenerateQrCodeConnectorAction` with a dedicated flag on each `site`?

I will start working on Site Areas.

LucasBrazi06:
> Thank you @LucasBrazi06 , this is a much cleaner and simpler approach! It keeps the component (i.e. Site) abstracted from specific permissions by just using the generic `readOnly` flag. I've been testing and didn't find issues so far. 🙂

Perfect :)

 > * If `canUpdate` already considers `issuer`, isn't `sites-list-table-data-source.ts:189` redundant?

I have nothing in line 189 in this file and in all the branches. But if you talk about the buildTableDynamicRowActions() method, then yes you are absolutely right, we can get rid of it 👍 !

> * Should we take into account `moreActions` such as `exportOCPPParamsAction` or `siteGenerateQrCodeConnectorAction` with a dedicated flag on each `site`?

I think yes, we can have canExportOCPPParams and canGenerateQrCode also but it has to be implemented first in the backend. I let you check this with @roxiciobotea210 .
 
> I will start working on Site Areas.

Perfect!
Thanks for your feedback!

PR #2377, opened 2021-04-23
Tags csv export front

cabch:
+ also changed ocpp local params export

### DIFF
--- a/src/assets/i18n/de.json
+++ b/src/assets/i18n/de.json
@@ -27,7 +27,7 @@
     "connectors": "Verknüpfungen",
     "notifications": "Benachrichtungen",
     "app_name": "e-Mobility",
-    "last_update": "Version 2.4.64",
+    "last_update": "Version 2.4.66",
     "version": "Version",
     "about_us": "About Us",
     "unexpected_error_backend": "Ein unerwarteter Fehler ist aufgetreten",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ev-dashboard",
-  "version": "2.4.64",
+  "version": "2.4.66",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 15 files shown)


src/app/pages/users/tags/tags-list-table-data-source.ts
```diff
@@ -314,6 +318,15 @@ export class TagsListTableDataSource extends TableDataSource<Tag> {
           (actionDef as TableImportTagsActionDef).action(ImportDialogComponent, this.dialog);
         }
         break;
+      case UserButtonAction.EXPORT_TAGS:
+        if (actionDef.action) {
+          const filterValues = this.buildFilterValues();
+          filterValues['WithUser'] = 'true';
```
LucasBrazi06, line 324: I don't think this filtering has to be done here.
Filtering of the list should be reused.
If extra params need to be forced then add them in the backend.

src/assets/i18n/de.json
```diff
@@ -1299,7 +1299,10 @@
         "description": "Dieser Nutzer wurde nicht mit dem Rechnungssystem synchronisiert.",
         "action": "Nutzer synchronisieren."
       }
-    }
+    },
+    "export_tags_title": "Export Tags",
+    "export_tags_confirm": "Do you really want to export the Tag list to a CSV format?",
+    "export_tags_error": "Error occurred while trying to export the Tags"
```
LucasBrazi06, line 1305: "Error occurred while trying to export the Tags, check the logs"

LucasBrazi06 (changes requested):
Small changes

src/app/pages/users/tags/tags-list-table-data-source.ts
cabch: I forgot to ask, will we let the choice to the user (checkbox or whatever) to chose if he wants to export with users or not ?

LucasBrazi06: For the time being let's always export all the tags with all the users and we'll see if the customers request this feature.

PR #2378, opened 2021-04-23
Avoid empty password on user creation

Ninjeneer:


src/app/pages/users/user/user.component.ts
```diff
@@ -234,11 +234,11 @@ export class UserComponent extends AbstractTabComponent implements OnInit {
           Validators.compose([
             Users.passwordWithNoSpace,
             Users.validatePassword,
-          ])),
+          ].concat(this.currentUserID && this.currentUserID !== '' ? [] : [Validators.required]))),
         repeatPassword: new FormControl('',
           Validators.compose([
             Users.validatePassword,
-          ])),
+          ].concat(this.currentUserID && this.currentUserID !== '' ? [] : [Validators.required]))),
```
jerome-benoit, line 241: Better is to test for `stringFoo?.length === 0`. 
There's no `isEmptyString()` or similar helper in Utils.ts?

jerome-benoit (changes requested):
Empty string test can be simplified.

src/app/pages/users/user/user.component.ts
Ninjeneer: > Better is to test for stringFoo?.length === 0.

stringFoo?.length === 0 doesn't give me the expected result
If the `currentUserID`is undefined, it will return me a `false` 

>  There's no isEmptyString() or similar helper in Utils.ts?

It seems not

jerome-benoit: You're right. Let's keep the test as it is.

jerome-benoit (approved):

jerome-benoit (approved):

jerome-benoit:
@LucasBrazi06: same remark as for the mirror PR on ev-server.

PR #2380, opened 2021-04-23
Use charging station id to initialize limitation dialog

OliveGerste:


LucasBrazi06 (approved):
Thanks Tobias!
I introduced this regression, my bad!

PR #2392, opened 2021-05-05
Authorizations > Front-end: Enhancement of the current concept (Site Areas)

Fonsu:


src/app/pages/organization/site-areas/site-area/site-area.component.html
```diff
@@ -12,7 +12,7 @@
               <div class="row app-organization">
                 <div class="col-md-5 d-block mt-auto">
                   <div class="rotating-card-container"
-                    [ngStyle]="{ 'pointer-events': canCreateSiteArea || canUpdateSiteArea ? 'all' : 'none'}">
+                    [ngStyle]="{ 'pointer-events': readOnly ? 'all' : 'none'}">
```
roxiciobotea210, line 15: Is the new check equivalent to the old one? if so, shouldn't it be `readOnly ? 'none' : 'all'` ? Or `!readOnly ? 'all' : 'none'` as it is used below (line 24)?

src/app/types/Authorization.ts
```diff
@@ -164,6 +164,13 @@ export interface SiteAuthorizationActions extends AuthorizationActions {
   canUnassignUsers?: boolean;
 }
 
+export interface SiteAreaAuthorizationActions extends AuthorizationActions {
+  canAssignChargingStations?: boolean;
+  canUnassignChargingStations?: boolean;
+  canAssignAssets?: boolean;
+  canUnassignAssets?: boolean;
+}
+
```
roxiciobotea210, line 173: This is where `exportOCPPParamsAction` and `siteAreaGenerateQrCodeConnectorAction` should also go in the future?

src/app/types/DataResult.ts
```diff
@@ -80,6 +81,11 @@ export interface SiteDataResult extends DataResult<Site> {
   canAssignUsers: boolean;
   canUnassignUsers: boolean;
 }
+export interface SiteAreaDataResult extends DataResult<SiteArea> {
+  canCreate: boolean;
+  canAssignChargingStations: boolean;
+  canUnassignChargingStations: boolean;
+}
```
roxiciobotea210, line 88: Doesn't it also need `canAssignAssets` & `canUnassignAssets`?

src/app/pages/organization/site-areas/list/site-areas-list-table-data-source.ts
```diff
@@ -203,99 +178,90 @@ export class SiteAreasListTableDataSource extends TableDataSource<SiteArea> {
...
-        ];
-        if (this.isAssetComponentActive) {
-          actions.splice(2, 0, this.assignAssetsToSiteAreaAction);
-        }
+    const moreActions = new TableMoreAction([]);
+    if (siteArea.canUpdate) {
+      actions.push(this.editAction);
+      // flag to be added on   the backend for increased granularity
+      moreActions.addActionInMoreActions(this.exportOCPPParamsAction);
+      moreActions.addActionInMoreActions(this.siteAreaGenerateQrCodeConnectorAction);
```
roxiciobotea210, line 245: Are these always available when canUpdate is true? I'm not sure at the moment what the backend check should be based on.

src/app/types/DataResult.ts
Fonsu: Actually that shouldn't be there, at the root level we only have `canCreate`. Will clean up `SiteDataResult` as well. 👍

src/app/pages/organization/site-areas/list/site-areas-list-table-data-source.ts
Fonsu: Yes, but it's only temporary. This is the way it was before the new auth concept implementation. Once the flag's in place it'll be driven by it.

src/app/types/Authorization.ts
Fonsu: Well, maybe it'd make more sense on `AuthorizationActions` since it's a common action.

src/app/pages/organization/site-areas/site-area/site-area.component.html
Fonsu: Thanks! That check's for the animation only but I missed the `!` 😅

jerome-benoit:
@roxiciobotea210: Could you please flag that as PR mergeable or not?

roxiciobotea210 (changes requested):
@Fonsu Let's keep it in "requested changes" status until we both add the 2 missing flags (`exportOCPPParamsAction` & `siteGenerateQrCodeConnectorAction`).

@jerome-benoit The 2 flags I mentioned also need to be added on BE side. They will be added in the next few days on both BE & FE, we'll test again after that and see if we're able to merge everything. 

Thanks!

PR #2396, opened 2021-05-06
Charging Profiles - Fixed power slider binding

OliveGerste:


src/app/pages/charging-stations/charging-station-limitation/cell-components/charging-station-power-slider-cell.component.ts
```diff
@@ -10,7 +10,7 @@ import { Utils } from '../../../../utils/Utils';
     <div class="row m-0">
       <app-charging-station-power-slider class="col-md-12"
         [chargingStation]="chargingStation" [chargePoint]="chargePoint"
-        [forChargingProfile]="true" [currentAmp]='row?.limit' (silderChanged)="sliderChanged($event)">
+        [forChargingProfile]="true" [currentAmp]='row?.limit' (sliderChangedEmitter)="sliderChanged($event)">
```
jerome-benoit, line 13: That should fix it, but there's no generic component event emitter to use instead to listen to the parent component events? 
That would then avoid to introduce regression because of a renaming.

jerome-benoit (approved):

PR #2400, opened 2021-05-11
Fill currentuserID with loggedUserID in profile edit

mahdibenromdhane:


src/app/pages/users/user/user.component.ts
```diff
@@ -149,6 +149,7 @@ export class UserComponent extends AbstractTabComponent implements OnInit {
     }
     this.isBillingComponentActive = this.componentService.isActive(TenantComponents.BILLING);
     this.canListPaymentMethods = this.authorizationService.canListPaymentMethods();
+    this.windowService.getPath() === '/users/profile' ? this.currentUserID = this.centralServerService.getLoggedUser().id : '';
```
jerome-benoit, line 152: I do not really like the hardcoding of the URI path to init a user identifier. 
Is it possible to trigger the init by using native angular routing or component name?

jerome-benoit: After more thoughts, I think first you need to understand why `currentUserID` attribute have became required to enable the save button when an admin edit a user or basic user edit his own profile. And if the requirement must be kept, the condition init the attribute must not depend on an URI path taken from the windowService, it's a security issue.

mahdibenromdhane: > After more thoughts, I think first you need to understand why `currentUserID` attribute have became required to enable the save button when an admin edit a user or basic user edit his own profile. And if the requirement must be kept, the condition to init the attribute must not depend on an URI path taken from the windowService, it's a security issue.

The currentUserID has been added in a PR linked to [this issue](https://github.com/sap-labs-france/ev-server/issues/2374). It's to prevent the creation of a user without a password.

mahdibenromdhane: I couldn’t find other way to identify that the user.component is used from the profile and not from the (create/update).
I didn’t notice that we already use the activatedRoute.snapshot in the bottom of the ngOnInit() to fill the currentUserID if it's a profile update. I'v placed it before the the form init and it's working.

jerome-benoit: So it was initialisation order issue.

jerome-benoit (approved):

src/app/pages/users/user/user.component.ts
jerome-benoit: > So it was initialisation order issue. The save button was trigger only if the attribute `currentUserID` is filled?

mahdibenromdhane: Yes It's an initialisation order issue, we only allow empty passwords input if it's a user update or a profile update. if it's a user creation , the password is required. 
image

PR #2409, opened 2021-05-19
Consolidate version handling in package.json

jerome-benoit:
Signed-off-by: Jérôme Benoit <jerome.benoit@sap.com>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ev-dashboard",
-  "version": "2.4.67",
+  "version": "2.4.67-rc1",
   "lockfileVersion": 1,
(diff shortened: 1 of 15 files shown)


jerome-benoit:
> Approved. but i do not undersatnd the "-rc1" suffix added to the version.
> Do we have some best practises here?

During Serge leave, I want to follow the pre-release best practice for all QA deployments matching a pre-release version on GitHub: `npm version prerelease --preid rc` with a features freeze: only bug fixes are allowed.

Once the pre-release cycle is finished, the final version. 

That means that @cabch PR need to be finished, approved and merged before closing the feature merge windows end.

PR #2415, opened 2021-05-24
Add required visaulID property in tag import popup

mahdibenromdhane:


jerome-benoit (approved):
Mergeable for rc2.

@LucasBrazi06: Ok to merge it and deploy on QA with the rc2?

PR #2420, opened 2021-05-26
Authorizations > Front-end: Enhancement of the current concept (Cars)

Fonsu:
Refactor to adopt the new authorisation concept on Car Management and Car Catalogs.


jerome-benoit:
Could you please add a more accurate description? It related to Car entity as far as I know. 

Thanks.

LucasBrazi06 (approved):

PR #2423, opened 2021-05-28
secure ui when activating billing of transactions

cabch:


src/app/pages/settings-integration/billing/stripe/settings-stripe.component.html
```diff
@@ -5,9 +5,11 @@
     </div>
     <h3 class="card-title">{{'settings.billing.stripe.title' | translate}}</h3>
   </div>
-
   <div class="card-body">
     <div class="row mt-1">
+      <div *ngIf="transactionBillingActivated" class="col-md-6 deactivated-setting-message">
```
ClaudeROSSI, line 10: This has a side effect on the position of the first the url. Can you change it?

image

src/assets/i18n/de.json
```diff
@@ -1493,6 +1493,7 @@
     },
     "billing": {
       "description": "Das Abrechnungsinterface ermöglicht Abrechnungsfunktionen für den Ladestationenbetreiber.",
+      "deactivated_setting_message": "Billing of transactions is activated, you cannot change the billing information anymore",
```
ClaudeROSSI: the term transaction is confusing (and only used internally). The UI always uses the term "sessions" (or charging sessions).

I'll make a proposal with different terms.

ClaudeROSSI (changes requested):
small side effect on the position of the url (to stripe).
Can you change it? 

I pushed some changes to replace the term "transactions" by the term "charging sessions"

src/app/pages/settings-integration/billing/stripe/settings-stripe.component.html
cabch: done ✅

src/assets/i18n/de.json
cabch: ok 👍🏻

ClaudeROSSI (approved):

PR #2429, opened 2021-06-01
fix saving when billing form is disabled

cabch:


ClaudeROSSI (approved):
billing setting UI small bug fixed. 
Tested 
and no risks

PR #2439, opened 2021-06-07
Statistics > Cost per user doesn't change

RohanAntony:
Cost per user in chart label set to 0 when user has no pricing data.


LucasBrazi06 (approved):
Connect the PR with the issue (ZenHub plugin)
Put a meaningful description for the PR.
That's it :-)

PR #2440, opened 2021-06-08
Current version should not be taken from the package.json file which …

LucasBrazi06:
…will be bundled by the Angular build:

### DIFF
--- a/src/app/layouts/auth/auth-layout.component.ts
+++ b/src/app/layouts/auth/auth-layout.component.ts
@@ -1,20 +1,23 @@
 import { Component, ElementRef, OnInit } from '@angular/core';
 import { NavigationEnd, Router } from '@angular/router';
 import { filter } from 'rxjs/operators';
-
-import { version } from '../../../../package.json';
+import { UtilsService } from 'services/utils.service';
 
 @Component({
   selector: 'app-layout',
   templateUrl: './auth-layout.component.html',
 })
 export class AuthLayoutComponent implements OnInit {
-  public version: string = version;
+  public version: string;
   public mobileMenuVisible: any = 0;
   private toggleButton: any;
   private sidebarVisible: boolean;
 
-  public constructor(private router: Router, private element: ElementRef) {
+  public constructor(
+    private utilsService: UtilsService,
+    private router: Router, private element: ElementRef) {
+    // Set the version
+    this.version = this.utilsService.getCurrentVersion();
     this.sidebarVisible = false;
   }
 
--- a/src/app/release-notes/release-notes.component.ts
+++ b/src/app/release-notes/release-notes.component.ts
@@ -1,4 +1,5 @@
 import { Component } from '@angular/core';
+import { UtilsService } from 'services/utils.service';
 
 interface ReleaseNotes {
   version: string;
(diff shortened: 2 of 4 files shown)


jerome-benoit:
> On 8 Jun 2021, at 08:16, FABIANO Serge ***@***.***> wrote:
> 
> …will be bundled by the Angular build:
> 



The package.json is not bundled at all in the final build: the version string is just replaced in the bundled files by the version is in the package.json file.

Do:
npm run build:prod
cd dist
ls -R | grep package -> nothing is returned

The current way of centralising a version is the documented angular way of doing it and is compliant with any tool automatic software releasing.

--
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

LucasBrazi06:
> The package.json is not bundled at all in the final build: the version is just replaced in the bundled files by the version string in the package.json file.

Why it is appearing in the screenshot above then?

jerome-benoit:
> The package.json is not bundled at all in the final build: the version is just replaced in the bundled files by the version string in the package.json file.
> 
> Why it is appearing in the screenshot above then?
> 
> 


Dunno, report the bug to the software developer.

--
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

LucasBrazi06:
> Dunno, report the bug to the software developer.

I find it normal that an import of a dependency in the code is bundled by Angular.
To not bundle it, it should be in the 'asset' directory and loaded via HTTP.
Here it's part of the code.

jerome-benoit:
> Dunno, report the bug to the software developer.
> 
> I find it normal that an import of a dependency in the code is bundled by Angular.
> To not bundle it, it should be in the 'asset' directory and loaded via HTTP.
> Here it's part of the code.
> 


The import is on a field of a JSON file, TS replace the field by the value at transpilation. The tool you are using is buggy.

--
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

LucasBrazi06:
> The import is on a field of a JSON file, TS replace the field by the value at transpilation. The tool you are using is buggy.



> The import is on a field of a JSON file, TS replace the field by the value at transpilation. The tool you are using is buggy.

I don't think so.

jerome-benoit:
> 
> The import is on a field of a JSON file, TS replace the field by the value at transpilation. The tool you are using is buggy.
> 
> I don't think so.
> 
That's how TS transpiler deal with JSON field, it’s in the documentation and that’s why importing the version from the package.json is the documented angular way of centralising a version number.

Do in dist:

grep -R mochatest *

Output nothing, the file is not imported in any ways.


--
Jérôme Benoit aka fraggle
Piment Noir - https://piment-noir.org
OpenPGP Key ID : 27B535D3
Key fingerprint : B799 BBF6 8EC8 911B B8D7 CDBC C3B1 92C6 27B5 35D3

LucasBrazi06:
> grep -R mochatest * Output nothing, the file is not imported in any ways.

Ok, it's more relevant than the 'ls' command.
I think that both statements are true, we have a package.json bundled but only with the version property.
So I'm fine with it.
I close the PR.

jerome-benoit:
Well, I would have kept it with one small change: it offers an helper to properly get the version, the helper just have to be changed to return the version in package.json file.

PR #2441, opened 2021-06-08
Fix Issue #1694. Search field missing reset.

RohanAntony:
Search filter is a common filter across all pages. As a result the changes were done to the table component where this filter is implemented. 

The HTML changes are adding a reset cross like the dialog filter available for inputs prior to the search filter. Clicking this reset cross calls the 'resetSearchFilter()' which sets empties the element value as well as the sets the search value in datasource to empty value. The table component is then refreshed.

### DIFF
--- a/src/app/shared/table/table.component.html
+++ b/src/app/shared/table/table.component.html
@@ -140,6 +140,9 @@
         <div *ngIf="dataSource.isSearchEnabled" class="col search-form">
           <mat-form-field>
             <input appAutofocus autocomplete="off" #searchInput [placeholder]="searchPlaceholder | translate" matInput>
+            <button mat-icon-button matSuffix (click)="resetSearchFilter()" aria-label="Add">
+              <mat-icon>clear</mat-icon>
+            </button>
           </mat-form-field>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


src/app/shared/table/table.component.html
```diff
@@ -140,6 +140,9 @@
         <div *ngIf="dataSource.isSearchEnabled" class="col search-form">
           <mat-form-field>
             <input appAutofocus autocomplete="off" #searchInput [placeholder]="searchPlaceholder | translate" matInput>
+            <button mat-icon-button matSuffix (click)="resetSearchFilter()" aria-label="Add">
```
OliveGerste, line 143: I think the aria-label does not really match the purpose of the button.

OliveGerste (changes requested):

src/app/shared/table/table.component.html
RohanAntony: True but if you look at line number 113 and line number 121 which both refer to the same button but for different elements, the aria-label used is "Add". I copied it to keep the format.

RohanAntony: I can change it to "Clear".

OliveGerste: Yes I think this is also not correct. You can change the aria-labels there too.

LucasBrazi06 (approved):

PR #2448, opened 2021-06-11
HOTFIX - billing endpoints - properly type the response

ClaudeROSSI:
I introduced a change on the backend-side which had a side effect on the deletion of a payment method.

The reponse type was not typed properly. This has been fixed


ClaudeROSSI:
@AlixH , @cabch : let's make sure both the dashbord and the mobile app use the same approach

LucasBrazi06 (approved):

PR #2459, opened 2021-06-17
Changes made to keep selected filter values when search value changes.

RohanAntony:
Selected rows were not included in the actions list when search was used. Adding a filter to include selected rows prevents them from being deleted.


LucasBrazi06 (approved):
GUT!

PR #2461, opened 2021-06-17
Add new filters dependencines

mahdibenromdhane:


LucasBrazi06 (approved):
- Added default value for issuer filter in constructor
- Added Site filter to Charging Station's filter
- Removed ternary operator for Org filters
- Check auth before adding filter (canListUsers()...)

-> To be taken into account for the next PR

PR #2467, opened 2021-06-21
Format createdBy/lastChangeBy properties Ids

mahdibenromdhane:


src/app/shared/table/table-data-source.ts
```diff
@@ -449,6 +449,9 @@ export abstract class TableDataSource<T extends TableData> {
         this.setTotalNumberOfRecords(data.count);
         // Display only projected fields
         if (!Utils.isEmptyArray(data.projectedFields)) {
+          // Format createdBy/lastChangeBy properties Ids
+          data.projectedFields = data.projectedFields.map(projectedField =>
```
LucasBrazi06, line 453: Use a 'if' with an 'includes' statement to check the name of the field and split inside it.
You know how much I like several statements in one line ;-)

LucasBrazi06 (changes requested):

src/app/shared/table/table-data-source.ts
mahdibenromdhane: I'v changed it to an if statement in the map function to format the projectedFields array. The include is in the for of the the tableColumnsDef in the line after.

LucasBrazi06 (approved):

PR #2471, opened 2021-06-23
Billing Settings - UI - clear test data

ClaudeROSSI:
FE : enable the "Delete Test Data" feature when targetting a test account


src/app/pages/settings-integration/billing/settings-billing.component.html
```diff
@@ -12,7 +12,7 @@ <h3 class="card-title">{{'settings.information' | translate}}</h3>
 </div>
 <form *ngIf="isActive" class="form" [formGroup]="formGroup">
   <div class="mat-toolbar-row left-actions">
-    <button mat-raised-button color="warn" (click)="save(formGroup.getRawValue())"
+    <button mat-raised-button color="warn" (click)="saveNewSettings(formGroup.getRawValue())"
```
LucasBrazi06, line 15: Why the renaming to saveNewSettings?

src/app/pages/settings-integration/billing/settings-billing.component.ts
```diff
@@ -69,17 +67,16 @@ export class SettingsBillingComponent implements OnInit {
...
+  public saveNewSettings(newSettings: any) {
     this.billingSettings.type = BillingSettingsType.STRIPE;
-    if (newSettings?.billing?.isTransactionBillingActivated) {
-      this.transactionBillingActivated = newSettings.billing.isTransactionBillingActivated;
-    } else {
-      this.transactionBillingActivated = this.billingSettings.billing.isTransactionBillingActivated;
-    }
-    this.billingSettings.billing = newSettings.billing as BillingSetting;
-    this.billingSettings.billing.isTransactionBillingActivated = this.transactionBillingActivated;
+    // Override current values with the new ones
```
LucasBrazi06, line 72: We should have the same approach (save, impl check)

src/app/pages/settings-integration/billing/settings-billing.component.html
```diff
@@ -28,6 +28,10 @@ <h3 class="card-title">{{'settings.information' | translate}}</h3>
       [disabled]="!formGroup.valid || formGroup.dirty || !billingSettings?.billing?.taxID || billingSettings?.billing?.isTransactionBillingActivated">
       <mat-icon>toggle_off</mat-icon><span>{{'settings.billing.transaction_billing_activation' | translate}}</span>
     </button>
+    <button *ngIf="billingSettings?.liveMode===false && billingSettings?.billing?.isTransactionBillingActivated" mat-raised-button color="warn" (click)="clearTestData()"
```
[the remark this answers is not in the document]
ClaudeROSSI: liveMode can be false, true or undefined.
I do not want to enable the "test data delation" when the type of the targeteed account could not be determined.

We can delete test data only when we are sure it's a test account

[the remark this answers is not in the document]
LucasBrazi06: Okay
WS remaining :-)

src/app/pages/settings-integration/billing/settings-billing.component.html
ClaudeROSSI: Well ... I can put back "save" as method name. 
But we have two different contexts when saving (because you asked us to avoid having a dedicated endpoint for the activation).
As a consequence, Caro was using save to "save" and to "activate" ... the result was confusing.
My personal way to do it woyld be to have "save" (for the regular part) and "_save" for the part which can be reused in both contexts ... but here again this is not something compliant with the 'spirit' of the other components.

So I do not know how to proceed here.
I personnaly also do not like the fact that the "save" input parameter is of type any.

src/app/pages/settings-integration/billing/settings-billing.component.ts
ClaudeROSSI: ok ... i'll have a look.
But our use-case is a bit different now that we have the STRIPE settings separated from the billing settings and dedicated endpoints with our billing router. We do not call the generic settings endpoint anynmore.

src/app/pages/settings-integration/billing/settings-billing.component.html
LucasBrazi06: Let's discuss this tomorrow around a coffee :)

ClaudeROSSI:
This PR mixes two different topics.
I'm closing this one and will do it with two new PR(s)

PR #2473, opened 2021-06-25
Csv import siteIDs front

cabch:
- added isImported to know where the user come from -> used in back
- added 'auto activation at import' to headers to send to BE + checkbox
- added siteIDs to optional properties in modal
- i18n messages to match Y/N dialog message


ClaudeROSSI (approved):
no comments from my side

LucasBrazi06 (changes requested):
I think we should add user's fields as optional fields in import Tags pop-up:

cabch:
import tag pop up optionnal fields modified 👌🏻

LucasBrazi06 (approved):

PR #2481, opened 2021-06-30
Rest endpoint - encode uri component

ClaudeROSSI:
Parameters that are part of the URL path are now encoded using encodeURIComponent (like this was done when using safe-url-assembler package)


ClaudeROSSI:
@AlixH : I think it makes sense to also do that change on the react-native side.
Please make sure to get rid of the dependency to 'safe-url-assembler' as well.

PR #2490, opened 2021-07-06
Billing - Invoice - Amount instead of Price as column title

ClaudeROSSI:


ClaudeROSSI:
Small change of the column title - Price replaced by **Amount**

PR #2510, opened 2021-07-27
Create cz

MichalAnthis:
Czech language mutation @mahdibenromdhane


ClaudeROSSI (changes requested):
Same comment than the ones made for the server side.

file name is not correct.
please format the content and check this is a valid json

PR #2513, opened 2021-07-28
Car connector - Mercedes framework + OAuth flow

OliveGerste:


LucasBrazi06 (approved):
We should probably stop stacking the code in the user component and start splitting it out :)
Let's see that in another PR.
I don't know if the user's consent will have to be in the auth module. I wanted to introduce a consent framework.
Maybe also the EULA will have to be aligned in the backend.

PR #2515, opened 2021-07-30
Data transfer   front

rmangin-fr:
Hi just to create the PR

### DIFF
--- a/src/app/services/central-server.service.ts
+++ b/src/app/services/central-server.service.ts
@@ -3109,6 +3109,10 @@ export class CentralServerService {
     return this.actionChargingStation(`${ServerRoute.REST_CHARGING_STATIONS}/${id}/cache/clear`, '');
   }
 
+  public chargingStationTriggerDataTransfer(id: string): Observable<ActionResponse> {
+    return this.actionChargingStation(`${ServerRoute.REST_CHARGING_STATIONS}/${id}/dataTransfer`, '');
+  }
+
   public actionChargingStation(action: string, args: string): Observable<ActionResponse> {
     // Verify init
     this.checkInit();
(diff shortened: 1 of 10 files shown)


src/app/services/central-server.service.ts
```diff
@@ -3109,6 +3109,10 @@ export class CentralServerService {
     return this.actionChargingStation(`${ServerRoute.REST_CHARGING_STATIONS}/${id}/cache/clear`, '');
   }
 
+  public chargingStationTriggerDataTransfer(id: string): Observable<ActionResponse> {
+    return this.actionChargingStation(`${ServerRoute.REST_CHARGING_STATIONS}/${id}/dataTransfer`, '');
```
ClaudeROSSI, line 3113: I think the best practise for REST endpoints is to avoid uppercase in routes.
I would suggest /data/transfer instead of /dataTransfer.
Of course the server as to be updated accordingly

src/assets/i18n/en.json
```diff
@@ -801,6 +801,7 @@
     "reboot_action": "Reboot",
     "soft_reset_action": "Soft Reset",
     "clear_cache_action": "Clear cache",
+    "trigger_data_transfer_action": "Trigger Data Transfer",
     "smart_charging_action": "Charge Limitation",
```
ClaudeROSSI, line 805: cz.json has been added in the meantime.
please update it as well when merging the latest master-qa branch. Thanks

ClaudeROSSI (changes requested):
REST route minor change needed

src/app/services/central-server.service.ts
LucasBrazi06: I agree, and as it's one command, to call it data-transfer as the others

ClaudeROSSI (approved):
sounds ok to me

ClaudeROSSI:
let's wait for the server-side and them mark this draft PR as "ready for review"

LucasBrazi06:
As discussed, not valid to have it in the front-end (vendor specific)
This will be used via Swagger.

PR #2543, opened 2021-09-02
Display warning if public charging stations exist under site about to be made private

RohanAntony:
QA: Setup a public site with public charging station and verify if an error is thrown when the site is made private without setting the charging station private.


RohanAntony:
@LucasBrazi06 If a charging station is made public when the site is private, an error is thrown by the backend but not being handled in the front-end yet. Raising an [issue](https://github.com/sap-labs-france/ev-dashboard/issues/2544).

src/app/pages/organization/sites/site/site.component.ts
```diff
@@ -288,6 +288,9 @@ export class SiteComponent implements OnInit {
         case HTTPError.OBJECT_DOES_NOT_EXIST_ERROR:
           this.messageService.showErrorMessage('sites.site_not_found');
           break;
+        case HTTPError.SITE_WITH_PUBLIC_CHARGER_ERROR:
```
LucasBrazi06, line 291: Only one name for Charging Station, we avoid now CHARGER or ChargeBox...

LucasBrazi06 (changes requested):
Also check public Charging Station on non public Site

LucasBrazi06 (approved):

PR #2551, opened 2021-09-07
REST > Updated assets routes to REST API

Ninjeneer:


src/app/utils/Utils.ts
```diff
@@ -925,4 +925,13 @@ export class Utils {
     return typeof value === 'string' ? '"' + value.replace(/"/g, '""') + '"' : value;
   }
 
+  public static removeUnsetAttributes(data: any) {
+    // Remove every undefined, null or empty strings attributes from an object to avoid triggering backend validation with wrong values
+    for (const key of Object.keys(data)) {
```
LucasBrazi06, line 930: Dangerous as we sometimes want to nullify a property.
It should work without this and null props should be explicitly set.
What do you think?

LucasBrazi06 (changes requested):

src/app/utils/Utils.ts
Ninjeneer: Yep you're right, I'll nullify the values causing problems

LucasBrazi06 (approved):

PR #2553, opened 2021-09-07
Tag tab ,basic user manage his own badges , user not mandatory in badges anymore

mahdibenromdhane:


LucasBrazi06:
@mahdibenromdhane Why the name of your branch is daf28b1 !? :-)

mahdibenromdhane:
> @mahdibenromdhane Why the name of your branch is [daf28b1](https://github.com/sap-labs-france/ev-dashboard/commit/daf28b15a6ec5093dc80bf6b8dfdd52e0db34f39) !? :-)
I Had the same problem with git push that we faced yesterday in the server, so I generated a branch from a previous commit and the branch took the commit id as a name

src/app/pages/transactions/history/transactions-history-table-data-source.ts
```diff
@@ -126,12 +126,12 @@ export class TransactionsHistoryTableDataSource extends TableDataSource<Transact
       this.loadUserFilterLabel(userID);
     }
     // Tag
-    const tagID = this.windowService.getSearch('TagID');
-    if (tagID) {
+    const visualTagID = this.windowService.getSearch('VisualTagID');
```
LucasBrazi06, line 129: VisualTagID -> VisualID ?

LucasBrazi06 (commented):
Just small comments.
I'll test the UI now.

LucasBrazi06 (changes requested):
- What happens if an active unassigned badge is used to start a transaction?
- If not Assigned to a User you cannot set the Default flag (UI + backend) -> Force default to false
- For Basic user rename the Assign/Unassign (only i18n):
    - Assign -> Register
    - Unassign -> Delete
- Unassign in table row -> Use the same icon and color as in the toolbar button
- When basic user 'deletes' the tag, it should be marked as inactive
- Basic user cannot assign an inactive badge to himself
- He cannot change the status of the badge either

mahdibenromdhane:
> * What happens if an active unassigned badge is used to start a transaction?
> * If not Assigned to a User you cannot set the Default flag (UI + backend) -> Force default to false
> * For Basic user rename the Assign/Unassign (only i18n):
>   
>   * Assign -> Register
>   * Unassign -> Delete
> * Unassign in table row -> Use the same icon and color as in the toolbar button
> * When basic user 'deletes' the tag, it should be marked as inactive
> * Basic user cannot assign an inactive badge to himself
> * He cannot change the status of the badge either

- an error occur if an active unassigned badge is used to start a transaction
- image
- Done for the rest of remarks

src/app/pages/tags/tag-assign/tag-assign.component.ts
```diff
@@ -77,12 +77,10 @@ export class TagAssignComponent implements OnInit {
     this.userID = this.formGroup.controls['userID'];
     this.default = this.formGroup.controls['default'];
     this.default.setValue(false);
-    if (!this.canListUsers) {
```
mahdibenromdhane: Removed because siteOwner can list users but cannot assign/create badge for other users in authorizationsDefinition

LucasBrazi06 (approved):
I approve but change the following:
- UserID is mandatory for Site Admin when creating a Badge

PR #2565, opened 2021-09-13
Technical users

rmangin-fr:


src/app/types/User.ts
```diff
@@ -40,6 +40,8 @@ export interface User extends TableData, CreatedUpdatedProps {
   eulaAcceptedVersion: number;
   eulaAcceptedOn: Date;
   billingData: BillingUserData;
+  technical: boolean;
+  projectedFields: string[];
```
LucasBrazi06, line 44: projectedFields can be put in the global type named TableData which will be used everywhere

LucasBrazi06 (changes requested):

src/app/types/User.ts
rmangin-fr: you mean the global type TableData defined in Table.ts (line 22):
export interface **TableData** extends AuthorizationActions {
?

LucasBrazi06: AuthorizationActions is even better indeed :-)

PR #2615, opened 2021-09-30
Filter technical users

rmangin-fr:


LucasBrazi06 (approved):
I had to update the MongoDB request in the backend:
```
    // Select (non) technical users
    if (Utils.objectHasProperty(params, 'technical') && Utils.isBoolean(params.technical)) {
      if (params.technical) {
        filters.technical = true;
      } else {
        const technicalFilter = [
          { technical: { $in: [false, null] } },
          { technical: { $exists: false } }
        ];
        if (filters.$or) {
          filters.$or.push(
            ...technicalFilter
          );
        } else {
          filters.$or = technicalFilter;
        }
      }
    }
```

PR #2626, opened 2021-10-03
Prevent xss injections

TheGodMother25:


src/app/shared/table/table.component.html
```diff
@@ -296,9 +297,8 @@
             <ng-container *ngIf="!dataSource.tableDef.rowDetails.angularComponent">
               <td class="col-3em details-column-row"></td>
               <td [attr.colspan]="numberOfColumns-1">
-                <div
-                  [innerHtml]="(dataSource.tableDef.rowDetails ? row[dataSource.tableDef.rowDetails.detailsField] : '')"
-                  class="row-detail">
+                <div class="row-detail">
+                  {{(dataSource.tableDef.rowDetails ? row[dataSource.tableDef.rowDetails.detailsField] : '')}}
```
TheGodMother25, line 301: Side effect: the formatting of the detailed logs will be lost as the output is not longer interpreted.
If we set it back to **innerHtml** then in case we have HTML that's in the stack trace (error coming from an external service which is saved in the logs), the HTML will be executed and the output will look like this
image

LucasBrazi06: I would need the formatted logs as I cannot read them anymore:

src/app/authentication/eula/authentication-eula.component.html
```diff
@@ -3,7 +3,9 @@
     <div class="container">
       <div class="row">
         <div class="col-md-12 ml-auto mr-auto mt-5 pt-5 eula-page">
-          <div [innerHTML]="eulaText" class="lighter-link-color"></div>
+          <div class="lighter-link-color">
+            {{eulaText}}
```
LucasBrazi06, line 7: EULA is not interpreted anymore.

LucasBrazi06 (changes requested):

src/app/authentication/eula/authentication-eula.component.html
TheGodMother25: here innerHTML could be used as soon as the eula is trusted source :)
I'll do the change

src/app/shared/table/table.component.html
TheGodMother25: indeed this the side effect, we can use the innerHTML but the html could be interpreted as above :(

src/app/authentication/eula/authentication-eula.component.html
LucasBrazi06: Yes, indeed, we are the only ones that can change it, delivered in the repo.
No endpoint can change this content.

src/app/shared/table/table.component.html
LucasBrazi06: Yes, security sometimes make our life harder.
If you can find a solution to format the text without having security issue, feel free :)

TheGodMother25: the formatting is back, to avoid HTML rendering i'll try to escape these characters "< >"

src/app/authentication/eula/authentication-eula.component.html
TheGodMother25: Fixed ;)

src/app/shared/table/table.component.html
LucasBrazi06: Great, thank you!!!

TheGodMother25:
> Here: 
> 
> If you don't have it, try to install the ZenHub Chrome extension :-)

Done ;)

PR #2746, opened 2021-11-16
Only disable autoUserSiteAssignment when property exist and value to false

mahdibenromdhane:
…false

### DIFF
--- a/src/app/pages/organization/sites/site/site.component.ts
+++ b/src/app/pages/organization/sites/site/site.component.ts
@@ -121,7 +121,7 @@ export class SiteComponent implements OnInit {
         if (site.address) {
           this.address = site.address;
         }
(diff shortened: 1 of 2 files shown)


src/app/pages/organization/sites/site/site.component.ts
```diff
@@ -121,7 +121,7 @@ export class SiteComponent implements OnInit {
         if (site.address) {
           this.address = site.address;
         }
-        if (!site.metadata?.autoUserSiteAssignment.enabled) {
+        if (site.metadata?.autoUserSiteAssignment && site.metadata?.autoUserSiteAssignment.enabled === false) {
```
LucasBrazi06, line 124: I don't like the === false in the if ;-)

LucasBrazi06 (changes requested):

PR #2752, opened 2021-11-18
Frontend - australian locale

ClaudeROSSI:
Adding the Australian locale en_AU required to reverse the logic for resolving the current locale.

- with the former implementation, the locale was determine from the language. This was not possible with the australian use-case since the langauge is "en"
- the new logic does the opposite - we now make sure to first determine the locale and then the language.

I also find that the cz_CZ locale does not exists:
- https://lh.2xlibre.net/locale/cs_CZ/
We should have been susing cs_CZ from the beginning
- https://lh.2xlibre.net/locale/cs_CZ/

The registration of the locales has been fixed in the app.module.ts

Screenshot from 2021-11-18 17-28-55


LucasBrazi06 (approved):
GUT!!!

PR #2847, opened 2021-12-15
Pricing restrictions UI

cabch:
without time picker for the first shot
still some comments to be checked for second phase


src/assets/i18n/cs.json
```diff
@@ -159,6 +159,20 @@
...
+    "friday": "friday",
+    "saturday": "saturday",
+    "sunday": "sunday",
+    "monday_short": "mon",
+    "tuesday_short": "tue",
+    "wednesday_short": "wed",
+    "thursday_short": "thu",
+    "friday_short": "fri",
+    "saturday_short": "sat",
+    "sunday_short": "sun",
```
ClaudeROSSI, line 175: I think in english the weekdays short from should be Capitalized and end with a dot
- https://www.englishclub.com/vocabulary/time-days-of-week.htm

src/assets/i18n/de.json
```diff
@@ -159,6 +159,20 @@
...
+    "friday": "friday",
+    "saturday": "saturday",
+    "sunday": "sunday",
+    "monday_short": "mon",
+    "tuesday_short": "tue",
+    "wednesday_short": "wed",
+    "thursday_short": "thu",
+    "friday_short": "fri",
+    "saturday_short": "sat",
+    "sunday_short": "sun",
```
ClaudeROSSI, line 175: idem

src/assets/i18n/en.json
```diff
@@ -159,6 +159,20 @@
...
+    "friday": "friday",
+    "saturday": "saturday",
+    "sunday": "sunday",
+    "monday_short": "mon",
+    "tuesday_short": "tue",
+    "wednesday_short": "wed",
+    "thursday_short": "thu",
+    "friday_short": "fri",
+    "saturday_short": "sat",
+    "sunday_short": "sun",
```
ClaudeROSSI, line 175: idem

src/assets/i18n/fr.json
```diff
@@ -159,6 +159,20 @@
...
+    "friday": "vendredi",
+    "saturday": "samedi",
+    "sunday": "dimanche",
+    "monday_short": "lun",
+    "tuesday_short": "mar",
+    "wednesday_short": "mer",
+    "thursday_short": "jeu",
+    "friday_short": "ven",
+    "saturday_short": "sam",
+    "sunday_short": "dim",
```
ClaudeROSSI, line 175: in french. I would expect 3 letters and a dot. - and it's Ok for having the first letter lowercase

ClaudeROSSI (changes requested):
some comments on the texts!

General remark for the UI. can we reduce the space between two dimensions to have a shorter screen heigth?

src/assets/i18n/cs.json
LucasBrazi06: Have you tried to retrieve it from moment()?
https://momentjscom.readthedocs.io/en/latest/moment/06-i18n/07-listing-months-weekdays/

ClaudeROSSI: > Have you tried to retrieve it from moment()? https://momentjscom.readthedocs.io/en/latest/moment/06-i18n/07-listing-months-weekdays/

Be aware that, on the server-side, we use isoWeekday() to have monday as the first day of the week.
>>> 1:MONDAY, 7: SUNDAY

c.f.: method: ConsumptionPricer.**checkDayOfWeek**() ...

I think, that we will have to take into account the current locale to determine which day as to be the first "toggle button" (of the week) but this is not urgent.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -94,17 +119,42 @@ export class PricingDefinitionComponent implements OnInit {
...
     private router: Router,
-    public translateService: TranslateService) {
+    public translateService: TranslateService,
+    public dayPipe: AppDayPipe) {
   }
 
   public ngOnInit(): void {
-    // TODO : show current entity name instead of id - for others than charging station c'est pas relevant
+    // Need to set locale manually depending on loggedUser locale for days abbreviation displaying
+    moment.locale(this.centralServerService.getLoggedUser().locale);
```
ClaudeROSSI, line 128: This has an impact on all UIs and all controls. As a consequence, setting the moment.locale should not be done here but in a central place. We already have issues with the datepicker control which does not consider the user locale right now. We need to initialize it only once and as soon as the user logs on. We also need to make sure we never change it afterward.

src/assets/i18n/cs.json
```diff
@@ -1207,13 +1207,14 @@
...
-    "locale_desc_italian": "Italština",
-    "locale_desc_czech": "Czech",
+    "locale_desc_en_US": "Angličtina",
+    "locale_desc_fr_FR": "Francouzština",
+    "locale_desc_es_ES": "Španělština",
+    "locale_desc_de_DE": "Němčina",
+    "locale_desc_pt_PT": "Portugalština",
+    "locale_desc_it_IT": "Italština",
+    "locale_desc_cs_CZ": "Czech",
+    "locale_desc_en_AU": "Australian",
```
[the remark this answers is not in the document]
cabch: Oui, j’ai aligné avec le fichier en.json car j’ai vu que ce n’était pas bon en changeant les clés des Days

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -326,26 +445,54 @@ export class PricingDefinitionComponent implements OnInit {
     }
   }
 
+  private clearAndResetControl(control: AbstractControl) {
+    control.reset();
+    control.clearValidators();
+    control.updateValueAndValidity();
+  }
+
   private consistencyCheck(pricingDefinition: PricingDefinition) {
```
ClaudeROSSI, line 454: The name of the method is confusing, because the code changes the pricingDefinition object passed as input.
I woyld expect a method to check the consistency and a method to get rid of properties that have been switched OFF.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -326,26 +445,54 @@ export class PricingDefinitionComponent implements OnInit {
...
-    }
-    for (const dimensionKey in pricingDefinition.dimensions) {
-      if (!pricingDefinition.dimensions[dimensionKey].active) {
-        delete pricingDefinition.dimensions[dimensionKey].price;
+    if (this.timeFromEnabled.value) {
+      if (this.timeToValue.value < this.timeFromValue.value) {
+        this.messageService.showErrorMessage('settings.pricing.pricing_definition_end_time_error');
+        this.timeToValue.setErrors({incorrectEndTime: true});
+        this.timeToValue.markAsPristine();
+        return false;
```
ClaudeROSSI, line 460: I think we should be able to define a tariff which is valid from 23:00 to 08:00.
The check done here is therefore not necessary.

The only situation which is incorrect is when the time range is from 08:00 to 08:00 !

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -326,26 +445,54 @@ export class PricingDefinitionComponent implements OnInit {
...
-    if (!this.chargingTimeEnabled.value || !this.chargingTimeStepEnabled.value) {
-      delete pricingDefinition.dimensions.chargingTime.stepSize;
-    }
-    if (!this.energyEnabled.value || !this.energyStepEnabled.value) {
-      delete pricingDefinition.dimensions.energy.stepSize;
-    }
-    for (const dimensionKey in pricingDefinition.dimensions) {
-      if (!pricingDefinition.dimensions[dimensionKey].active) {
-        delete pricingDefinition.dimensions[dimensionKey].price;
+    if (this.timeFromEnabled.value) {
```
ClaudeROSSI, line 455: no else? ... returns undefined. We cannot create a tariff with no time restrictions anymore!
Please split the method logic in two as suggested in my previous comment!

ClaudeROSSI (changes requested):
checkConsistency is too strict and does not allow creating a simple tariff anymore

src/app/types/Pricing.ts
```diff
@@ -68,25 +68,16 @@ export interface PricingStaticRestriction {
 
 export interface PricingRestriction {
   // TODO - first Pricing implementation - some options are not yet supported
-  // startDate?: string, // Start date, for example: 2015-12-24, valid from this day
-  // endDate?: string, // End date, for example: 2015-12-27, valid until this day (excluding this day)
-  // startTime?: string, // Start time of day, for example 13:30, valid from this time of the day. Must be in 24h format with leading zeros. Hour/Minute se
-  // endTime?: string, // End time of day, for example 19:45, valid until this time of the day. Same syntax as start_time
-  // daysOfWeek?: DayOfWeek[], // Which day(s) of the week this tariff is valid
+  // startDate?: string; // Start date, for example: 2015-12-24, valid from this day
+  // endDate?: string; // End date, for example: 2015-12-27, valid until this day (excluding this day)
```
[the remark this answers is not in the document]
cabch: has been changed

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: need to make some change in the entire application, it has been removed for now

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: changed to checkPricingDefinition is it ok for you @ClaudeROSSI ?

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: changed to show an error if times are the same (timeFrom === timeTo) is it ok for you @ClaudeROSSI ?

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: see above modif

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
```diff
@@ -252,6 +252,132 @@
...
               </div>
             </div>
+            <!-- Restrictions title -->
+            <div class="row">
+              <div class="col-md-5 text-left">
+                <span class="pricing-header">
+                  {{'settings.pricing.restrictions_title' | translate}}
+                </span>
+              </div>
+            </div>
```
ClaudeROSSI, line 262: Can we put the title inside the card? just for a better look and feel?
We have a card body section. do we have a card header section?

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -315,6 +421,19 @@ export class PricingDefinitionComponent implements OnInit {
     this.minDate = moment(event.value).toDate();
   }
 
+  public setMinTime(event) {
+    this.minTime = event.currentTarget.value;
+  }
```
ClaudeROSSI, line 426: Why do we have setMinTime() and we do not have set MaxTime()???

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -315,6 +421,19 @@ export class PricingDefinitionComponent implements OnInit {
...
+
+  private setTimeFormat() {
+    // if there is AM/PM in longDateFormat means it's 12hours time format
+    if (moment.localeData().longDateFormat('LT').toLowerCase().indexOf('a') > -1) {
+      this.timeFormat = '12';
+    } else {
+      this.timeFormat = '24';
```
ClaudeROSSI, line 433: Is this necessary? 
We need to clarify if this works when the moment locale gets changed dynamically (taking into account the user locale after logon - another PR is ongoing to have the js locale and the moment locale in sync!

src/assets/scss/views/pricing/pricing.scss
```diff
@@ -31,3 +31,7 @@
 .pricing-not-set {
   color: $state-danger-bg;
 }
+
+.days-of-week-button {
+  width: 4em;
```
ClaudeROSSI, line 36: Can we use min-width instead of width to make sure the button size is always sufficient (in case the translated texts are longer than the english ones).

ClaudeROSSI (approved):
Approved!

But still a few comments

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -35,6 +36,7 @@ export class PricingDefinitionComponent implements OnInit {
   public formGroup!: FormGroup;
   public currentPricingDefinition: PricingDefinition;
   public context: string;
+  public timeFormat: string;
```
ClaudeROSSI, line 39: timeFormat is set but not used!

ClaudeROSSI (approved):
I did a minor change (fr.json)
Approved!

Still some comments.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: was for handling 12-24hours format, will remove it

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: was to handle maxTime not to be before minTime, not used anymore, will get rid of it

cabch: same for minDate that is already handled with the datepicker

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: will delete

src/assets/scss/views/pricing/pricing.scss
cabch: we can, but translated text will always be 3 letters as it's for short day name, still relevant to change to min-width ?

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
cabch: here is the before Screenshot 2022-01-03 at 12 41 35
after
Screenshot 2022-01-03 at 12 40 23
tell me if it's what you have think of ?

cabch: there is a card header section, this is what shows the "toto - chargngstation" name at the top of the above screenshot

src/assets/scss/views/pricing/pricing.scss
ClaudeROSSI: 3 letters! This is true for the language we have so far. But we don't know for 'exitic' languages! - And eve if it's 3 letters, we do not know the size of each letters. do we?

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
ClaudeROSSI: ok. I prefer the second screenshot!

src/assets/scss/views/pricing/pricing.scss
cabch: we don't :) will push changes

ClaudeROSSI (approved):
ok. Thanks

ClaudeROSSI:
@serge: I think we are ready for the UI now.
Only thing missing is the handling of the extra inactivities (backend side)

LucasBrazi06:
> @serge: I think we are ready for the UI now. Only thing missing is the handling of the extra inactivities (backend side)

You have time, I don't expect reviewing this PR until this Friday.

ClaudeROSSI:
> > @serge: I think we are ready for the UI now. Only thing missing is the handling of the extra inactivities (backend side)
> 
> You have time, I don't expect reviewing this PR until this Friday.

For the extra inactivity, we already have the problem in prod (because the previous Ui already incudes the parking time option). So we may consider it as a bug!

This PR is independent and, as such, can be reviewed/merged as soon as possible but with its backend counter part (i.e.: price consumption chunks).

LucasBrazi06:
PR with no issue!

ClaudeROSSI:
> PR with no issue!

ok i'll create one now

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
ClaudeROSSI: ok

ClaudeROSSI:
> > PR with no issue!
> 
> ok i'll create one now

Done: 
https://github.com/sap-labs-france/ev-server/issues/3228

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
```diff
@@ -252,6 +252,132 @@
...
+                    </mat-slide-toggle>
+                  </div>
+                  <!-- start time -->
+                  <div class="col-md-2">
+                    <mat-form-field *ngIf="timeFromEnabled.value">
+                      <mat-label>{{'settings.pricing.restriction_start_time' | translate}}</mat-label>
+                      <input matInput type="time" id="timeFromValue" name="timeFromValue" [formControl]="timeFromValue" (change)="setMinTime($event)" required/>
+                    </mat-form-field>
+                    <mat-error *ngIf="timeFromValue?.errors?.required">{{'general.mandatory_field' | translate}}</mat-error>
+                    <mat-error *ngIf='timeToValue?.errors?.incorrectEndTime'>{{'settings.pricing.pricing_definition_time_error' | translate}}</mat-error>
```
LucasBrazi06, line 350: Why not incorrectStartTime here ?

LucasBrazi06: And I don't know when this is displayed as you change it to the right value after entering an out of range time.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
```diff
@@ -252,6 +252,132 @@
...
+                  </div>
+                  <!-- end time -->
+                  <div class="col-md-2">
+                    <mat-form-field *ngIf="timeFromEnabled.value">
+                      <mat-label>{{'settings.pricing.restriction_end_time' | translate}}</mat-label>
+                      <input matInput type="time" id="timeToValue" name="timeToValue" [formControl]="timeToValue" required/>
+                    </mat-form-field>
+                    <mat-error *ngIf="timeToValue?.errors?.required">{{'general.mandatory_field' | translate}}</mat-error>
+                    <mat-error *ngIf='timeToValue?.errors?.incorrectEndTime'>{{'settings.pricing.pricing_definition_time_error' | translate}}</mat-error>
```
LucasBrazi06, line 359: There is no wrong end time now as it spans to the next day.

LucasBrazi06, line 359: When you have the same start/end time, I assume it means 1 minute?

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
```diff
@@ -252,6 +252,132 @@
...
+                <div class="row">
+                  <!-- toggle days -->
+                  <div class="col-md-2 pricing-dimension">
+                    <mat-slide-toggle (change)="toggle($event)" id="daysOfWeek" [checked]="daysOfWeekEnabled.value">
+                      {{'settings.pricing.restriction_days_of_week' | translate}}
+                    </mat-slide-toggle>
+                  </div>
+                  <!-- value days -->
+                  <div class="col-md-2">
+                    <mat-button-toggle-group *ngIf="daysOfWeekEnabled.value" [value]="selectedDays" [formControl]="selectedDays" name="daysOfWeekToggle" aria-label="Days of week" multiple>
```
LucasBrazi06, line 372: Use the primary color for the day selection.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -243,24 +321,50 @@ export class PricingDefinitionComponent implements OnInit {
...
+    if (this.checkPricingDefinition(pricingDefinition)) {
+      if (this.currentPricingDefinitionID) {
+        this.updatePricingDefinition(pricingDefinition);
+      } else {
+        this.createPricingDefinition(pricingDefinition);
+      }
     }
   }
 
   public toggle(event) {
```
LucasBrazi06, line 333: Use specific toggle method: toggleTimeFrom()...
And type the event param according the control type.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -326,7 +426,21 @@ export class PricingDefinitionComponent implements OnInit {
...
-  private consistencyCheck(pricingDefinition: PricingDefinition) {
+  private clearAndResetControl(control: AbstractControl) {
+    control.reset();
+    control.clearValidators();
+    control.updateValueAndValidity();
+  }
+
+  private checkPricingDefinition(pricingDefinition: PricingDefinition) {
+    if (this.timeFromEnabled.value) {
+      if (this.timeToValue.value === this.timeFromValue.value) {
```
LucasBrazi06, line 437: Should be displayed in the UI, check how password is handled.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
```diff
@@ -344,8 +458,21 @@ export class PricingDefinitionComponent implements OnInit {
...
     }
+    if (this.daysOfWeekEnabled.value) {
+      pricingDefinition.restrictions.daysOfWeek = this.selectedDays.value;
+    }
+    if (!this.timeFromEnabled) {
+      delete pricingDefinition.restrictions.timeFrom;
+      delete pricingDefinition.restrictions.timeTo;
+    }
+    for (const restrictionKey in pricingDefinition.restrictions) {
+      if (!pricingDefinition.restrictions[`${restrictionKey}`]) {
```
LucasBrazi06, line 469: Remove the quote and use restrictionKey right away.

LucasBrazi06: I don't get the '!' in the if statement: if there is no restriction then you delete it!?

LucasBrazi06 (changes requested):
Check my comments.
Charging Time is still in EUR/Hour! After user's feedback, change it to EUR/Min

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
cabch: I display the 'timeRangeError' error message now

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
cabch: you're right, changed it to 'timeRangeError' to display "same from/end time" error (if timeFromValue === timeEndValue)

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: it's displayed when we save already

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: we remove property when it's null or 0 because back end it;s optimized : property is not set by default

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
LucasBrazi06: The error message should be displayed in the UI and not when you click on save.
But I think you did it already in your latest changes, right?

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
LucasBrazi06: Ok, thanks.

ClaudeROSSI: > Remove the quote and use restrictionKey right away?

@cabch : Serge is right the **`${}`** syntax here is use less. Just remove it!

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
ClaudeROSSI: > Check my comments. Charging Time is still in EUR/Hour! After user's feedback, change it to EUR/Min

@LucasBrazi06 : EUR/Hour vs EUR/Min ??? we have to discuss here. The best would be to have some options here. Both should be accepted.
If we look at the tariff created by Kinza on SLF we have: 5EUR/hour, with a step size of 10 minutes.
This would be **0.08333333333** EUR/minutes ==> and thus **0.8333333333** EUR/Steps.
I do not see the benefit for the end-user to enter it per minutes. I would suggest to keep the current entry and just display additional information next to it (just to help the end-user).

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
LucasBrazi06: Okay, it would be good to add something like included/excluded in the labels.

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
LucasBrazi06: This was the first feedbacks from Gerald.
Having it per hour and receiving the Meter Values and pricing them per min would almost always result in a floating point.
Furthermore, the meter value interval can be changed (one every 5 mins).
Maybe a drop-down with 'mins' and 'hours' would be the solution :-).

ClaudeROSSI: > 

ok but in a the next PR lol

LucasBrazi06: > > 
> 
> ok but in a the next PR lol

Ca fait la deuxieme fois que t'essaies de me la faire a l'envers lol
If it's just a front-end change do it right now ;-P

ClaudeROSSI: > > > 
> > 
> > 
> > ok but in a the next PR lol
> 
> Ca fait la deuxieme fois que t'essaies de me la faire a l'envers lol If it's just a front-end change do it right now ;-P

I think this is not so simple. Where do we store the fact that the user entered it per minutes and not in hours?
If you switch from one unit to the other (on a browser refresh) it will just make the UI eve more confusing.
I need customer feedbacks ... the feeling from Gerald is not sufficient. Especially when I see the mess that we have in our own Gireve Tariff definitions.

LucasBrazi06: > I need customer feedbacks ... the feeling from Gerald is not sufficient. Especially when I see the mess that we have in our own Gireve Tariff definitions.

Even Kinza expected mins and got lost and this was also a feeling I had.

> I think this is not so simple. Where do we store the fact that the user entered it per minutes and not in hours?

This was my initial question, if it requires a change in the backend we can postpone it but not forever ;-).

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: yes I changed it in other places but not here, change has been done

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.ts
cabch: done

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
cabch: I added it in the label is that what you meant ?
Screenshot 2022-01-17 at 09 29 38

src/app/shared/formatters/app-currency.pipe.ts
```diff
@@ -7,24 +7,26 @@ import { LocaleService } from '../../services/locale.service';
...
     this.localeService.getCurrentLocaleSubject().subscribe((locale) => {
       this.currencyPipe = new CurrencyPipe(locale.currentLocaleJS);
     });
-    // Get the Pricing settings
-    this.currency = this.centralServerService.getLoggedUser().currency;
+    // Get the currency code
+    // The currency code is part of the user token and comes from the Pricing Settings!
+    this.centralServerService.getCurrentUserSubject().subscribe((user) => {
+      // user is null when logging out
+      this.currencyCode = user?.currency;
```
ClaudeROSSI, line 22: @LucasBrazi06 : Pipe has been changed to make sure we get the latest **currency code** from the **userToken**.

We subscribe here to get notified when the user **logs in** and **logs out**. With this approach we make sure that each amount / price is formatted using the correct currency.

However, when changing the Pricing Settings, a logout **has to be forced** to make sure we get a fresh **userToken**. This is not yet done in this PR!

LucasBrazi06: I don't see why we need to subscribe here?

As discussed, why not using an Angular service to pull that value from the backend.
I don't know if we have similar use case with other variables.

And as discussed with Caroline, I didn't want to handle the logout from the backend.

Maybe the simplest solution would be that when the user changes the currency + click on Save -> Display a warning message that tells that it will be logged out + Customer accepts + Call backend to save the currency + Force logout.

ClaudeROSSI: > I don't see why we need to subscribe here?
> 
> As discussed, why not using an Angular service to pull that value from the backend. I don't know if we have similar use case with other variables.
> 
> And as discussed with Caroline, I didn't want to handle the logout from the backend.
> 
> Maybe the simplest solution would be that when the user changes the currency + click on Save -> Display a warning message that tells that it will be logged out + Customer accepts + Call backend to save the currency + Force logout.

With the previous code, the logout had no impact. The pipe instance was not recreated after a new logon and the previous currency code was still used. This is the reason why have changed the logic.

Now, each **AppCurrencyPipe** gets the up-to-date currency code (from the user token) at log on. This is fast enough because the token is parsed only once and kept in the **centralServerService**.

LucasBrazi06: > With the previous code, the logout had no impact. The pipe instance was not recreated after a new logon and the previous currency code was still used. This is the reason why have changed the logic.

Have you checked how Angular handles pipe and how dependency injection can influence the lifetime of it?

> Now, each **AppCurrencyPipe** gets the up-to-date currency code (from the user token) at log on. This is fast enough because the token is parsed only once and kept in the **centralServerService**.

Okay, but to get the new token, you'll need to logout the user, right?

ClaudeROSSI: > > With the previous code, the logout had no impact. The pipe instance was not recreated after a new logon and the previous currency code was still used. This is the reason why have changed the logic.
> 
> Have you checked how Angular handles pipe and how dependency injection can influence the lifetime of it?
> 
> > Now, each **AppCurrencyPipe** gets the up-to-date currency code (from the user token) at log on. This is fast enough because the token is parsed only once and kept in the **centralServerService**.
> 
> Okay, but to get the new token, you'll need to logout the user, right?

right. The logout is necessary ... at least, now with a logout we see the new currency.
Before we had to refresh the browser.

ClaudeROSSI: > > > With the previous code, the logout had no impact. The pipe instance was not recreated after a new logon and the previous currency code was still used. This is the reason why have changed the logic.
> > 
> > 
> > Have you checked how Angular handles pipe and how dependency injection can influence the lifetime of it?
> > > Now, each **AppCurrencyPipe** gets the up-to-date currency code (from the user token) at log on. This is fast enough because the token is parsed only once and kept in the **centralServerService**.
> > 
> > 
> > Okay, but to get the new token, you'll need to logout the user, right?
> 
> right. The logout is necessary ... at least, now with a logout we see the new currency. Before we had to refresh the browser.

I have a question: Why would you subscribe for the localeJs and not for the currency code? Both info comes from the token. So we should have the same logic for both. Do you agree?

LucasBrazi06: > I have a question: Why would you subscribe for the localeJs and not for the currency code? Both info comes from the token. So we should have the same logic for both. Do you agree?

It's not because I made a choice 2 years ago that it's still the right one today after 4 major Angular releases ;-).
Why do we cache the currency in the pipe, why don't we always use the centralServerService to get the currency?
What would be the overhead?

ClaudeROSSI: @LucasBrazi06 : As far as i understand, the pipe lifecycle is bound to component life-cycle itself.
Based on the test I did, our components are not destroyed when the user logs out. This is the reason why the subscribe() is the only option with the current flow.
BTW: I also need to change pipe: **app-pricing-dimensions-unit** - this one displays the currency code (CAN) instead of the symbol ($CA)

LucasBrazi06: Remove the class members and always call the service?
Did you test this?

This would remove also the locale subscription.

ClaudeROSSI: just received your reply. 
I'll  give it a try

ClaudeROSSI:
@LucasBrazi06 , @cabch 

FYI: I changed the UI to show minutes instead of seconds in some of the input fields

ClaudeROSSI:
@LucasBrazi06 : I sent you a mail about some changes to make sure the currency code cannot be changed in the pricing settings too easily.
Let me know what you think.
I'll stop changing that PR now!!! ;)

src/app/pages/settings-integration/pricing/simple/settings-simple-pricing.component.html
```diff
@@ -11,13 +11,15 @@ <h3 class="card-title">{{'settings.pricing.settings.title' | translate}}</h3>
       <div class="col-md-6">
         <mat-form-field>
           <mat-label for="inputPriceUnit">{{'pricing.price_unit_label' | translate}} *</mat-label>
-          <mat-select id="inputPriceUnit" [formControl]="currency">
+          <mat-select id="inputPriceUnit"
+            [disabled]="isCurrencyCodeReadonly"
```
LucasBrazi06, line 15: Violation of golden rule 36:

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
```diff
@@ -9,7 +9,7 @@
           </ng-template>
```
LucasBrazi06, line 9: Missing padding/margin at the bottom:

src/app/shared/pricing-definitions/pricing-definition/pricing-definition.component.html
```diff
@@ -252,11 +252,137 @@
                 </div>
               </div>
             </div>
+            <!-- form Restrictions -->
+            <div class="card">
```
LucasBrazi06, line 256: I can enter min duration which is greater than max duration: no error message is shown
Same for Min Energy.
When day of week is selected, the text should be white (poor contrast with black)

LucasBrazi06 (approved):
Check my remarks.
I'll merge this PR for testing it in QA but comment/fix my feedbacks.
Thanks!

PR #2872, opened 2021-12-28
Asset auth migration

Ramzay:
Hello,
This PR is about migrating the Asset component to ABAC.
There is another one for the server part ;)
Enjoy !


src/app/pages/assets/asset/asset.component.ts
```diff
@@ -191,13 +181,21 @@ export class AssetComponent implements OnInit {
   }
 
   public loadAsset() {
+    // ID not provided we are in creation mode
     if (!this.currentAssetID) {
+      if (this.authorizationService.canCreateAsset()) {
```
mahdibenromdhane, line 186: I think there is no need for this check (we are trying to get rid of the authorizationService on the frontend).

src/app/pages/assets/asset/asset.component.ts
```diff
@@ -495,4 +493,9 @@ export class AssetComponent implements OnInit {
       }
     });
   }
+
+  private enableForm(): void {
```
mahdibenromdhane, line 497: This function should be replaced with: 
 this.readOnly = (this.dialogMode === DialogMode.VIEW);
 like It's used in CompanyComponent, SiteComponent ..

src/app/services/authorization.service.ts
```diff
@@ -96,6 +96,10 @@ export class AuthorizationService {
     return this.canAccess(Entity.ASSET, Action.UPDATE);
   }
 
+  public canCreateAsset(): boolean {
```
mahdibenromdhane, line 99: No need for this function since we are getting the create auth flag in the list.

mahdibenromdhane (changes requested):

src/app/pages/assets/asset/asset.component.ts
Ramzay: If the plan is to remove the authorizationService then I agree, but in that case we might need to send a flag to the pop-up that the user is allowed to create an asset (when we retrieve the asset list)

mahdibenromdhane: > If the plan is to remove the authorizationService then I agree, but in that case we might need to send a flag to the pop-up that the user is allowed to create an asset (when we retrieve the asset list)

I'm not sure if this going to be necessary if we display the create button only if the user is allowed to create (which is case)

src/app/pages/assets/in-error/assets-in-error-table-data-source.ts
```diff
@@ -48,13 +47,11 @@ export class AssetsInErrorTableDataSource extends TableDataSource<AssetInError>
...
     private componentService: ComponentService,
     private centralServerService: CentralServerService,
-    private authorizationService: AuthorizationService,
     private dialog: MatDialog,
     private dialogService: DialogService) {
     super(spinnerService, translateService);
     // Init
-    this.isAdmin = this.authorizationService.isAdmin();
-    this.setStaticFilters([{WithSiteArea: true}]);
+    this.setStaticFilters([{ WithSiteArea: true }]);
```
LucasBrazi06, line 53: In line 145: Should we check canListSite/SiteArea to display the filters?

src/app/pages/assets/list/assets-list-table-data-source.ts
```diff
@@ -175,32 +173,39 @@ export class AssetsListTableDataSource extends TableDataSource<Asset> {
...
+    if (!Utils.isEmptyArray(moreActions.getActionsInMoreActions())) {
+      moreActions.addActionInMoreActions(openInMaps);
+      rowActions.push(moreActions.getActionDef());
+    } else {
+      // More action is empty we put actions directly in row
       rowActions.push(openInMaps);
     }
     // Display refresh button
-    if (this.isAdmin && asset.dynamicAsset && !asset.usesPushAPI) {
+    if (asset.canRetrieveConsumption) {
```
LucasBrazi06, line 208: No splice, push it at the right place.

LucasBrazi06 (approved):
Check my comments
I'll merge as I refactored a bit the file namings ;-)

PR #2894, opened 2022-01-05
Quick option to set a commonly used date range in to and from field.

RohanAntony:
Test
- Go to Logs page and click on the timer icon. 
- Check each dropdown value to verify if the From and To is being selected as well as dates in logs output is within range

Added an option in filter that looks like a calendar with event marked on it.
image

When selecting this week
image

### DIFF
--- a/src/app/shared/table/table.component.html
+++ b/src/app/shared/table/table.component.html
@@ -94,8 +94,10 @@
             [class]="(filterDef.class ? filterDef.class : '')">
             <!-- Button based Dropdown filter -->
             <div *ngIf="filterDef.type === FilterType.MENU_DROPDOWN">
-              <button mat-icon-button [matMenuTriggerFor]="menu">
-                <mat-icon>{{ filterDef.label ? filterDef.label : 'more_vert' }}</mat-icon>
+              <button mat-stroked-button [matMenuTriggerFor]="menu">
+                <button mat-icon-button [disableRipple]='true'>
+                  <mat-icon>{{ filterDef.label ? filterDef.label : 'more_vert' }}</mat-icon>
+                </button>
               </button>
               <mat-menu #menu="matMenu">
                 <div *ngFor="let item of filterDef.items">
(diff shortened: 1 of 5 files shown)


LucasBrazi06:
@RohanAntony Quick feedback.
I would have put the selection after the date filters and choose an icon like a stop watch to not confuse it with the ones in the filters.
Thanks!

RohanAntony:
> @RohanAntony Quick feedback. I would have put the selection after the date filters and choose an icon like a stop watch to not confuse it with the ones in the filters. Thanks!

image

LucasBrazi06:
> image

In white theme :-) and is there a way to display it in a button?

RohanAntony:
> > image
> 
> In white theme :-) and is there a way to display it in a button?

image

src/app/shared/table/table.component.ts
```diff
@@ -128,8 +129,30 @@ export class TableComponent implements OnInit, AfterViewInit, OnDestroy {
     }
   }
 
-  public filterChanged(filterDef: TableFilterDef) {
-    this.dataSource.filterChanged(filterDef);
+  public filterChanged(filterDef: TableFilterDef, key?: string) {
+    if(filterDef.id === 'date_menu_dropdown') {
```
LucasBrazi06, line 133: This code should not be there.
The two date filters should be passed to the DateMenuDropdownFilter object which will update the Json date objects according the drop down selection.
The this code should be in the DateMenuDropdownFilter.
I don't know if the framework is supporting that currently.
You can check with Mahdi who did the dependencies between filters.

src/app/shared/table/table.component.html
```diff
@@ -92,6 +92,21 @@
         <ng-template ngFor let-filterDef [ngForOf]="dataSource.tableFiltersDef">
           <div *ngIf="filterDef.visible !== false"
             [class]="(filterDef.class ? filterDef.class : '')">
+            <!-- Button based Dropdown filter -->
+            <div *ngIf="filterDef.type === FilterType.MENU_DROPDOWN">
+              <button mat-stroked-button [matMenuTriggerFor]="menu">
```
LucasBrazi06, line 97: Use the primary (blue) button design (same as the ones in the toolbar)

LucasBrazi06, line 97: If it's possible also, reduce the width of the button so we have a square button and not a rectangular one.

LucasBrazi06 (changes requested):
Check my comments

RohanAntony:
Closing PR as [Date Range Picker PR](https://github.com/sap-labs-france/ev-dashboard/pull/2895) replaces this.

PR #2895, opened 2022-01-10
Using DateTimeRangePicker with custom ranges instead of independent From, To and menu option to select custom ranges.

RohanAntony:
Test:
- Check if the datetime range picker works in Logs page as expected.
- Check if the date ranges shown in the datetime range picker selects the right start and end datetime values.
- Check if all the text involved is translated to respective language when locale is changed for all languages.
- Check if reset filters works as expected: reset startdate: 12:00 AM and endDate: now
- Check the same functionality for all the below pages
  - Sessions > History
  - Sessions > In Error
  - Invoices
  - Statistics (all tabs)

Current output (Updated image)
image


RohanAntony:
If we add " position:fixed" CSS to the injected  element, the output is as follow. I am not sure as to how we can accomplish this. 
image
image

src/app/shared/table/table.component.ts
```diff
@@ -25,6 +26,7 @@ import { TableDataSource } from './table-data-source';
 export class TableComponent implements OnInit, AfterViewInit, OnDestroy {
   @Input() public dataSource!: TableDataSource<TableData>;
   @ViewChild('searchInput') public searchInput!: ElementRef;
+  @ViewChild(DaterangepickerDirective) public picker: DaterangepickerDirective;
```
LucasBrazi06, line 29: If you have 2 date time range picker control in the template based on different props (start/endDate) how do you handle them with one picker?

src/app/shared/table/table.component.ts
```diff
@@ -134,6 +136,29 @@ export class TableComponent implements OnInit, AfterViewInit, OnDestroy {
     this.refresh();
   }
 
+  public dateRangeChanged(filterDef: TableFilterDef, event: any){
```
LucasBrazi06, line 139: [low] White space (WS) before the curly brace (install eslint in VSCode)

src/app/shared/table/table.component.ts
```diff
@@ -134,6 +136,29 @@ export class TableComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    const currentValue = filterDef.currentValue;
+    if(currentValue?.startDate !== event.startDate || currentValue?.endDate !== event.endDate){
```
LucasBrazi06, line 141: [low] WS.
Check everywhere.

src/app/shared/table/table.component.ts
```diff
@@ -134,6 +136,29 @@ export class TableComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  public createQuickAccessDateRanges(rangeObjects: {key: string; label: string; startValue: Date; endValue: Date}[]){
+    const dateRanges: any = {};
+    for (const range of rangeObjects) {
+      dateRanges[this.translateService.instant(range.label)] = [range.startValue, range.endValue];
+    }
+    return dateRanges;
+  }
+
+  public openDateRange() {
+    this.picker.open();
```
LucasBrazi06, line 159: Can't this be the default behavior without having to provide this code?

src/assets/css/dateTimeRangePickerDropdown.css
```diff
@@ -0,0 +1,4 @@
+/* CSS to fix the daterange dropdown picker issue */
+.md-drppicker.drops-down-auto.ltr.shown.double.show-ranges {
```
LucasBrazi06, line 2: Could be put in src/assets/scss/views/shared/_date-picker.scss ? or a new file like _date_time_range_picker.scss?

LucasBrazi06 (changes requested):

src/app/shared/table/table.component.ts
RohanAntony: We can't handle it right now as both filters will have the same id. We update the filter values based on id. We need to make the id as an argument that can be modified during filter creation. Additional changes might also be required. As this piece of code is necessary for the calendar icon which on click must open the picker, I have removed the calendar icon. By default as the value is displayed, addition of a calendar icon might not be necessary.

src/app/shared/table/table.component.ts
RohanAntony: ESLint doesn't seem to be throwing errors for these on my machine.

src/app/shared/table/table.component.ts
RohanAntony: Removed the calendar icon. The [documentation](https://github.com/fetrarij/ngx-daterangepicker-material#open-datepicker-from-outside) doesn't provide an option to access the injected element within the DOM without this piece of code. We need to find a way to accomplish this.

src/app/shared/table/table.component.ts
LucasBrazi06: You can retrieve a list of elements in Angular:
https://angular.io/api/core/ViewChildren
Could you check?

RohanAntony: Tried with multiple inputs. It works. Removed the second element after trial.
image

src/app/shared/table/table-data-source.ts
```diff
@@ -345,6 +346,10 @@ export abstract class TableDataSource<T extends TableData> {
           if (filterDef.type === 'date') {
             // Date
             filterJson[filterDef.httpId] = filterDef.currentValue.toISOString();
+          } else if (filterDef.type === FilterType.DATE_RANGE) {
+            // Date range
+            filterJson['StartDateTime'] = filterDef.currentValue.startDate.toISOString();
+            filterJson['EndDateTime'] = filterDef.currentValue.endDate.toISOString();
```
LucasBrazi06, line 352: Still the same issue I think, if I have two date time range controls, you use the same properties to build the filter!
Filter properties should be provided when the filter is built and used here with the right httpIDs (2 props)

LucasBrazi06 (changes requested):
Check my comments.
And the locale is not working, check below in FR, I still have the time in AM/PM:

src/app/shared/table/table-data-source.ts
RohanAntony: I have added translation within the datetime picker for the apply button. We can also translate the month names and week names within the calendar but the display on the input will always show in English, I guess. Shall I add support for translating within the calendar?
image

LucasBrazi06: Normally you can do this with moment get this info.
I this Caroline or Claude got already this done, you can check with either of them.
Thanks!

LucasBrazi06 (changes requested):
In Statistics view the locale is not taken into account but is okay in Logs!?

Could you check?

LucasBrazi06 (approved):
Great job!

PR #2905, opened 2022-01-13
Use auth to show / hide filters

Ramzay:
- site and site areas filter visibility based on auth
- remove splice


LucasBrazi06 (approved):
GUT !

PR #2919, opened 2022-01-17
Use average per asset per minute before summing to site area level while displaying site area chart component

RohanAntony:
Test cases: 
- Check if the site area consumption chart aggregates per minute per asset before summing up to site area aggregation. 
  1. Manually add additional consumption to a site area asset for a given minute in Consumptions.
  2. Check if the average of the instantWatts is calculated for the given minute instead of summing the instantWatts within a minute.
- Check if the axis on site area consumption chart shows 'Consumption by Site Area'
- Check if the axis on site area consumption chart shows the values with 'A' or 'kW'
- Check if the negative axis is shown indicating production values.

Fixes:
- Uses non facet, only grouping based mongo db aggregation pipeline methods to aggregate.

Sample:
image

[ev-server changes](https://github.com/sap-labs-france/ev-server/pull/3245)


LucasBrazi06:
@RohanAntony Could you create an issue and link both PRs to it?
Thanks!

RohanAntony:
@LucasBrazi06 @OliveGerste For different assets under same site area, I manually cloned some production values and consumption values. The result is as below with the averaging being done as expected.
image

RohanAntony:
Y axis is now just 'Consumption' from a translated string instead of hardcoded string.

PR #2946, opened 2022-01-25
button style - text white when selected

ClaudeROSSI:
Here is the result!

image


ClaudeROSSI:
well ... i'm not sure if it's the right approach

PR #2958, opened 2022-01-27
Align car to new auth framework

Ramzay:
- Send authorisation object to the CREATE dialog: this allows to show/hide input forms based on authorisations
- DataResult now extends Auth object instead of adding auth attributes (only done for Car but can be extended to the other components -> will be done in other PR)
- Handle new auth flags


src/app/pages/cars/cars/cars-list-table-data-source.ts
```diff
@@ -226,19 +235,24 @@ export class CarsListTableDataSource extends TableDataSource<Car> {
...
-    const tableFilterDef: TableFilterDef[] = [];
-    if (this.authorizationService.canListUsers()) {
-      tableFilterDef.push(new UserTableFilter().getFilterDef());
-    }
-    tableFilterDef.push(new CarMakerTableFilter().getFilterDef());
-    return tableFilterDef;
+    this.usersFilter = new UserTableFilter().getFilterDef();
+    this.carMakerFilter = new CarMakerTableFilter().getFilterDef();
+    // Filter visibility will be defined by auth
+    this.usersFilter.visible = false;
```
LucasBrazi06, line 248: Visible should be set to false by default. No need to set it explicitly.

src/app/pages/cars/cars/cars-list-table-data-source.ts
```diff
@@ -64,7 +65,15 @@ export class CarsListTableDataSource extends TableDataSource<Car> {
...
-        this.createAction.visible = cars.canCreate;
+        // Initialize authorization actions
+        this.authorizationActions = {
+          canCreate: Utils.convertToBoolean(cars.canCreate),
+          canListUsers: Utils.convertToBoolean(cars.canListUsers),
+          canListCarCatalog: Utils.convertToBoolean(cars.canListCarCatalog),
+          canCreatePoolCar: Utils.convertToBoolean(cars.canCreatePoolCar)
+        };
+        // Update filters visibility
+        this.updateFilterVisibilityWithAthorizationActions();
```
LucasBrazi06, line 76: No need a method here I think or add the same kind of method in all the table data source :).

src/app/shared/table/actions/table-create-action.ts
```diff
@@ -24,7 +24,7 @@ export class TableCreateAction implements TableAction {
   }
 
   protected create(component: ComponentType<unknown>, dialog: MatDialog,
-    dialogParams: DialogParams<TableData> = {}, refresh?: () => Observable<void>, size?: PopupSize) {
+    dialogParams: DialogParamsWithAuth<TableData, AuthorizationActions> = {}, refresh?: () => Observable<void>, size?: PopupSize) {
```
LucasBrazi06, line 27: same remark as in Car but with TableData and AuthorizationActions.

src/app/shared/table/actions/cars/table-create-car-action.ts
```diff
@@ -22,8 +23,8 @@ export class TableCreateCarAction extends TableCreateAction {
   }
 
   private createCar(carDialogComponent: ComponentType<unknown>,
-    dialog: MatDialog, refresh?: () => Observable<void>) {
-    super.create(carDialogComponent, dialog, null, refresh, {
+    dialog: MatDialog, dialogParams: DialogParamsWithAuth<Car, CarAuthorizationActions>, refresh?: () => Observable<void>) {
```
LucasBrazi06, line 26: The goal to introduce AuthorizationActions was to have only one type of param for create dialog so I would have expected only this param: `dialogParams: AuthorizationActions`

src/app/pages/cars/cars/cars-list-table-data-source.ts
```diff
@@ -226,19 +235,24 @@ export class CarsListTableDataSource extends TableDataSource<Car> {
       case CarButtonAction.CREATE_CAR:
         if (actionDef.action) {
           (actionDef as TableCreateCarActionDef).action(CarDialogComponent,
-            this.dialog, this.refreshData.bind(this));
+            this.dialog, { authorizations: this.authorizationActions }, this.refreshData.bind(this));
```
LucasBrazi06, line 238: I would have expected only `this.authorizationActions` instead of `{ authorizations: this.authorizationActions }`
It gives the impression that you can pass the Car in addition to this param.

LucasBrazi06 (changes requested):

src/app/shared/table/actions/table-create-action.ts
Ramzay: Problem is that if I only put the AuthorizationActions we won't be able to pass metadata to the dialog if we needed to...   
With this current implementation If we need auth object we pass it along the metadata, ex User:

{ dialogData: { metadata: this.metadata } as User,  **authorizations: { ...authorizations }** }

To be compliant with your previous remarks, we would need to add an optional parameter to the create() and append the authorisations to the dialogParams like:

dialogConfig.data = {
dialogMode: DialogMode.CREATE,
 ...dialogParams **<- contains dialogData : {metadata: { ... }, ... }**
authorizations: { ...authorizations } **<- contains { canList : true }**
};

The **authorization:** would be important otherwise the dialogConfig.data would contain a blob and we would be forced to do something like this in car-dialog.component.ts

public constructor(dialogRef: MatDialogRef<CarDialogComponent>,
    @Inject(MAT_DIALOG_DATA) matDialogParams: **any**) {
    let dialogParams = matDialogParams as DialogParams<Car>;
    let carAuthorizationActions = matDialogParams as CarAuthorizationActions;
}

Let me know ! :)

src/app/pages/cars/cars/cars-list-table-data-source.ts
Ramzay: Ok !

src/app/shared/table/actions/table-create-action.ts
LucasBrazi06: > { dialogData: { metadata: this.metadata } as User, **authorizations: { ...authorizations }** }

We moved authorization out from the data object (Car, User...) for this purpose, why not metadata also :-)?
We can discuss this tomorrow if you are on site :-).

Ramzay:
Hello @LucasBrazi06

This PR is bigger than what I anticipated but the refactoring did that :) 
I separated entity and data-result authorisations, they are now distinct, so I had to create new types + clean unnecessary flags.

I kept the metadata as part of the list and the entity, I think it makes sense to have a general metadata at the LIST level and specific metadata at the Action level if needed:
- Admin: userID is not mandatory in the create action, but is mandatory for export action, 
- Basic: userID is always mandatory. 

Let me know regarding the metadata alignement, I will do it in the next PR when aligning the other components, CAR is the main focus here :)

src/app/types/Authorization.ts
```diff
@@ -157,22 +157,80 @@ export interface AuthorizationContext {
   assets?: string[];
 }
 
+export interface AuthorizationAttributes {
+  projectFields?: string[];
+  metadata?: Record<string, AuthorizationDefinitionFieldMetadata>;
+}
+// ENTITY: COMMON AUTH ACTION
 export interface AuthorizationActions {
   canRead?: boolean;
```
LucasBrazi06, line 166: canRead to remove in the next PR

LucasBrazi06 (approved):
Super !!!

PR #2976, opened 2022-02-08
Fix  set date range to 24 hours default

RohanAntony:
Datetime range filter is set to end of day instead of 'now()' preventing a require of updating date to for refresh.

Additionally
- 'Last minute', 'Last 10 minutes', 'Last 30 minutes' and 'Last Hour' values are updated every time the datetime range picker is opened. Previously, when page was kept open for more than a minute, "Date to" wouldn't be updated to latest minute. 
- Adding 'Last 24 hours' for ranges and other places where it might be useful as 'Last 24 hours' comes in between "Today" and "Yesterday".


LucasBrazi06 (approved):
GUT!

PR #2992, opened 2022-02-17
Fix screensizes part 4

cabch:


cabch:
N.B left to do :
- shorten list of width-levels / height-levels + breakpoint if needed
- uncomment

PR #3010, opened 2022-02-25
Fix for consumption chart which flickers as well as legend state being lost.

RohanAntony:
1. In sessions tab, select a consumption chart and check if only subset of data is being loaded initially. Data points for once every 3 minutes is loaded instead of once every minute. 
2.

### DIFF
--- a/src/app/shared/component/consumption-chart/consumption-chart.component.ts
+++ b/src/app/shared/component/consumption-chart/consumption-chart.component.ts
@@ -125,6 +125,7 @@ export class ConsumptionChartComponent implements AfterViewInit {
   public loadAllConsumptionsChanged(matCheckboxChange: MatCheckboxChange) {
     if (matCheckboxChange) {
       this.loadAllConsumptions = matCheckboxChange.checked;
+      this.updateVisibleDatasets();
       this.refresh();
     }
   }
@@ -489,9 +490,8 @@ export class ConsumptionChartComponent implements AfterViewInit {
         duration: 0,
       },
       responsive: true,
-      maintainAspectRatio: false,
       // spanGaps: true,
-      // aspectRatio: this.ratio,
+      aspectRatio: this.ratio,
       plugins: {
         legend: {
           position: 'bottom',
--- a/src/app/shared/table/table.component.ts
+++ b/src/app/shared/table/table.component.ts
@@ -149,6 +149,7 @@ export class TableComponent implements OnInit, AfterViewInit, OnDestroy {
       };
       for (const picker of this.datePickers) {
         picker.picker.updateCalendars();
+        picker.clear();
       }
       this.filterChanged(filterDef);
     }


ClaudeROSSI:
@RohanAntony : please put me in the loop for reviewing that PR.
@cabch : Caro, I've just tested that one, and we have vertical scrollbars when viewing the curves. Can you help here as well?

cabch:
updated with master and fixed dialog size @RohanAntony please tell me if you can see the dialog correctly without scroll bar nor big white space below the chart (you can try resizing your window manuelly, size should follow resizing)
@ClaudeROSSI : sounds fixed for my part

ClaudeROSSI:
> updated with master and fixed dialog size @RohanAntony please tell me if you can see the dialog correctly without scroll bar nor big white space below the chart (you can try resizing your window manuelly, size should follow resizing) @ClaudeROSSI : sounds fixed for my part

@cabch : I still have huge charts

image

RohanAntony:
> > updated with master and fixed dialog size @RohanAntony please tell me if you can see the dialog correctly without scroll bar nor big white space below the chart (you can try resizing your window manuelly, size should follow resizing) @ClaudeROSSI : sounds fixed for my part
> 
> @cabch : I still have huge charts
> 
> image

On line 539 in consumption-chart.component.ts, if you set "maintainAspectRatio: false", the chart is set to initial size but the flickering exists.

src/app/shared/component/consumption-chart/consumption-chart.component.ts
```diff
@@ -503,58 +550,19 @@ export class ConsumptionChartComponent implements AfterViewInit {
...
-              visibleDatasets.includes(ConsumptionChartDatasetOrder.INSTANT_VOLTS_DC)
-            ) {
-              this.gridDisplay[ConsumptionChartAxis.VOLTAGE] = true;
-            } else if( visibleDatasets.includes(ConsumptionChartDatasetOrder.STATE_OF_CHARGE) ) {
-              this.gridDisplay[ConsumptionChartAxis.PERCENTAGE] = true;
-            } else if( visibleDatasets.includes(ConsumptionChartDatasetOrder.CUMULATED_AMOUNT) ) {
-              this.gridDisplay[ConsumptionChartAxis.AMOUNT] = true;
-            }
+            this.updateVisibleGridLines();
             legend.chart.options = this.createOptions();
```
ClaudeROSSI, line 566: @RohanAntony : this line is suspicious to me. Why would you need to set the options again?
Chart options are to be set only once. right?

RohanAntony: We need to call createOptions again because the grid lines in the background are dynamically generated based on the visible legends. In scales option, we are setting the drawOnChartArea based on gridDisplay array values. This is updated when a legend is made visible or hidden. Technically, we might be able to update only the scales instead. 
To see the side effect on hiding this change, comment this line and open a sessions chart. Hide all fields that use (kW) or (kW.h). The background horizontal grid lines will disappear.

ClaudeROSSI:
> The flickering issue was due to the css value being set in % instead of vw or vh for the parent container. Also the default parameter for maintainAspectRatio in every chart is true which needs to be set to false inorder to fit the parent container. The size of parent container is set in _charts.scss file.
> 
> I have removed the default additional classes that were added to transaction.component.html. If we need this, we need to add custom media queries in the css file to handle when dialog is smaller else the chart overflows.

Let's sync on this.

src/app/shared/component/consumption-chart/consumption-chart.component.ts
```diff
@@ -658,102 +765,7 @@ export class ConsumptionChartComponent implements AfterViewInit {
...
-          ticks: {
-            callback: (value: number) => {
-              const result = this.appCurrencyPipe.transform(value, this.transaction.priceUnit);
-              return result ? result : '';
-            },
-            color: this.defaultColor,
-          },
-        }
-      },
+      scales: this.buildScales(),
```
ClaudeROSSI, line 768: @RohanAntony : I was expecting that the buildScales() method gets reused in the onClick callback when updating the legend. Did you face some unexpected problems?

You also forgot to remove the options private member which is not used anymore
>> private options!: ChartOptions;

ClaudeROSSI (changes requested):
@RohanAntony : have a look to my comments

src/app/shared/component/consumption-chart/consumption-chart.component.ts
RohanAntony: Lol, I forgot that. Will add it.

PR #3053, opened 2022-03-30
display TENANT_DEFAULT_LOGO on superTenant

mahdibenromdhane:


LucasBrazi06 (changes requested):
Check all the views in auth directory.
It's not only register and login I think.

mahdibenromdhane:
> Check all the views in auth directory. It's not only register and login I think.

Done

LucasBrazi06 (approved):
GUT Thanks !

PR #3063, opened 2022-04-06
Updating consumption chart to show updated legends on auto refresh.

RohanAntony:
Enable auto refresh on a charging session. Check if the state of any legend changed is maintained when chart gets refreshed.


ClaudeROSSI (approved):
tested and approved

PR #3092, opened 2022-04-21
Billing - fix - payment method spinner

ClaudeROSSI:
spinner now properly shown when adding a payment method

### DIFF
--- a/src/app/pages/users/user/payment-methods/payment-method/stripe/stripe-payment-method.component.ts
+++ b/src/app/pages/users/user/payment-methods/payment-method/stripe/stripe-payment-method.component.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/member-ordering */
 import { Component, ElementRef, Input, OnInit, ViewChild } from '@angular/core';
 import { AbstractControl, FormControl, FormGroup } from '@angular/forms';
 import { MatDialogRef } from '@angular/material/dialog';
@@ -21,9 +20,7 @@ import { PaymentMethodDialogComponent } from '../payment-method.dialog.component
   templateUrl: 'stripe-payment-method.component.html',
   styleUrls: ['stripe-payment-method.component.scss']
 })
-
 export class StripePaymentMethodComponent implements OnInit {
-
   @Input() public inDialog!: boolean;
   @Input() public dialogRef!: MatDialogRef<PaymentMethodDialogComponent>;
   @Input() public currentUserID!: string;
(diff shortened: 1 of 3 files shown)


src/app/pages/users/user/payment-methods/payment-method/stripe/stripe-payment-method.component.ts
```diff
@@ -143,72 +144,61 @@ export class StripePaymentMethodComponent implements OnInit {
...
+    return result;
   }
 
-  public close(saved: boolean = false) {
-    if (this.inDialog) {
-      this.dialogRef.close(saved);
+  private async attachPaymentMethod(operationResult: {setupIntent?: SetupIntent; error?: StripeError}) {
+    try {
+      this.spinnerService.show();
+      const response: BillingOperationResult = await this.centralServerService.setupPaymentMethod({
```
LucasBrazi06, line 194: Interesting technic to not call subscribe() and use async/await.
Should we generalize it?

LucasBrazi06: @ClaudeROSSI Any news?
In Angular in general we use subscribe (RxJs) not promises.

ClaudeROSSI: > @ClaudeROSSI Any news? In Angular in general we use subscribe (RxJs) not promises.

Well ... we need to clarify this. Is there any side-effects ... I do not yet know.

ClaudeROSSI: In RxJS 7 toPromise will become **deprecated** and with RxJS 8 - it will be gone! 

> https://indepth.dev/posts/1287/rxjs-heads-up-topromise-is-being-deprecated

But for now, I have no easy solution to get rid of it.
The code has not been changed - we use this approach for a while. I would like to keep it as it is right now and only apply the recommended solution as soon as we switch to RxJS 7. 
Is it ok with you?

ClaudeROSSI: @LucasBrazi06 : I do not see the benefit of writing this instead. Do you?

image

ClaudeROSSI: > > @LucasBrazi06 : I do not see the benefit of writing this instead. Do you?
> 
> Why we need a Promise here!? Is it a constraint from Stripe API?

This is just to make the code human-readable for the caller!  async / await are (most of the time) just syntactic sugar

LucasBrazi06: > > Why we need a Promise here!? Is it a constraint from Stripe API?
> 
> This is just to make the code human-readable for the caller! async / await are (most of the time) just syntactic sugar

Okay, then as it'll be deprecated, use RxJs ;-)

PR #3109, opened 2022-05-10
Bug fix  date range picker invalid date

RohanAntony:
- Using the start date and end date from the date picker object instead of parsing the input string and passing it.
- Translating the last 24 hours into french

NOTE: This option doesn't work in portugese language although.


RohanAntony:
This doesn't seem to work with the portugese version. The issue here lies with the datetimepicker. I need to find an alternative version for this but rest of the languages seem to be working fine.

PR #3172, opened 2022-06-07
Migrate charging station service abac

r0mux:


src/app/pages/charging-stations/list/charging-stations-list-table-data-source.ts
```diff
@@ -116,105 +137,93 @@ export class ChargingStationsListTableDataSource extends TableDataSource<Chargin
...
-          headerClass: 'd-none d-xl-table-cell col-20p',
-        },
-        {
-          id: 'siteArea.name',
-          name: 'site_areas.title',
-          class: 'd-none d-xl-table-cell col-20p',
-          headerClass: 'd-none d-xl-table-cell col-20p',
-        },
-      );
+    const columns: TableColumnDef[] = [];
```
LucasBrazi06, line 140: You can return right away the column def, no need an extra var.

LucasBrazi06 (commented):
Just one global remark before I review this PR.
Check that no 'if' statement is used in method buildTableColumnDefs() in Data Source. Use the visible property.
It seems already to be the case in this PR but must be checked in all Data Sources (maybe in another PR)
Thanks!

Ramzay:
> Just one global remark before I review this PR. Check that no 'if' statement is used in method buildTableColumnDefs() in Data Source. Use the visible property. It seems already to be the case in this PR but must be checked in all Data Sources (maybe in another PR) Thanks!

Yes it'

LucasBrazi06 (approved):
GUTGUTGUT!!!

PR #3177, opened 2022-06-09
Changes for handling chartjs upgrade

RohanAntony:
Chart js upgrade to 3.8 breaks due to change in font data type in 3.8


LucasBrazi06:
@RohanAntony I don't see changes in pkg.json.
What's the purpose of your changes!?

RohanAntony:
The idea was to first push these changes which works on both the old version and the new version. Then I let dependabot update the version with the alternate thread. I can change package.json but I am not sure if package.json and package-lock.json should be in sync.

PR #3216, opened 2022-06-28
Translation Update German

OliveGerste:


ClaudeROSSI (approved):
@OliveGerste : thanks a lot

PR #3231, opened 2022-07-06
Bp transfers follow up

ClaudeROSSI:


ClaudeROSSI:
already changed in another PR

PR #3263, opened 2022-07-14
Billing Platform - Assign Account to Company

RohanAntony:
[ev-server](https://github.com/sap-labs-france/ev-server/pull/3718) changes to populate the account data to display


RohanAntony:
Shall I add ev-server and ev-dashboard changes to display account user instead of account id while fetching the billing details for the company?

RohanAntony:
I have added the changes in the last commit to display account name instead of account id. I am not sure if I should hide the entire operation behind a 'withAccount' flag? 

image

PR #3267, opened 2022-07-19
Upgrade angular v14

mahdibenromdhane:


LucasBrazi06 (approved):
Okay for a first migration but Form should be typed in the future:
https://netbasal.com/typed-reactive-forms-in-angular-no-longer-a-type-dream-bf6982b0af28

PR #3334, opened 2022-08-09
Bug fix tenants logo cross origin

cabch:
need to check why the back url contains the port etc...


ClaudeROSSI (approved):
This is just a workaround for now.
The real solution would be to change how the  URL is generated server-side.

PR #3335, opened 2022-08-10
Add platform fee tax ID support in UI when billing platform is enabled

RohanAntony:
[ev-server](https://github.com/sap-labs-france/ev-server/pull/3783) changes
1. Enable billing platform and check if billing account tax rate can be set?
2. Set billing account tax rate and check if billing account component can be accessed?
3. Check if when billing platform is disabled, billing account tax rate is not visible in the JSON response.
4. Check if reset data clears the billing account tax rate?


ClaudeROSSI (changes requested):
@RohanAntony : that taxID is used. Don't remove it. Just add a new one!

PLease check my comment on the server PR!

ClaudeROSSI (approved):
ok

PR #3356, opened 2022-08-31
PRICING - Adding support to view all prices

RohanAntony:
[ev-server](https://github.com/sap-labs-france/ev-server/pull/3843) changes

User can now view all pricing components at the tenant level.

Test cases:
- Ensure view all and displaying pricing components are only available at Tenant level
- Ensure create option is disabled when viewing all components
- Ensure dialog is in view mode only and not edit mode when viewing the component

### DIFF
--- a/src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
+++ b/src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
@@ -70,7 +70,7 @@ export class PricingDefinitionsTableDataSource extends DialogTableDataSource<Pri
       this.context.entityType = this.defaultContext.entityType = entityType;
     }
     this.context.entityName = this.defaultContext.entityName = entityName;
-    //Force a table refresh
+    // Rebuild the table as the number of columns changed
     this.initDataSource(true);
   }
 
--- a/src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
+++ b/src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
@@ -4,7 +4,8 @@ import { Router } from '@angular/router';
 import { TranslateService } from '@ngx-translate/core';
 import { Observable } from 'rxjs';
 import { TablePricingOverviewAction } from 'shared/table/actions/pricing/table-pricing-overview-action';
-import { TableViewPricingDefinitionsAction, TableViewPricingDefinitionsActionDef } from 'shared/table/actions/table-view-pricing-definitions-action';
+import { TableViewAction } from 'shared/table/actions/table-view-action';
+import { TableViewPricingDefinitionsActionDef } from 'shared/table/actions/table-view-pricing-definitions-action';
 import { DialogMode } from 'types/Authorization';
 import { ButtonAction } from 'types/GlobalType';
 
@@ -33,16 +34,16 @@ export class PricingDefinitionsTableDataSource extends DialogTableDataSource<Pri
   private createAction = new TableCreatePricingDefinitionAction().getActionDef();
   private editAction = new TableEditPricingDefinitionAction().getActionDef();
   private deleteAction = new TableDeletePricingDefinitionAction().getActionDef();
-  private viewAction = new TableViewPricingDefinitionsAction().getActionDef();
+  private viewAction = new TableViewAction().getActionDef();
   private defaultContext = {
-    entityID: '',
-    entityType: '',
-    entityName: ''
+    entityID: null,
+    entityType: null,
+    entityName: null
   };
   private context = {
-    entityID: '',
-    entityType: '',
-    entityName: ''
+    entityID: null,
+    entityType: null,
+    entityName: null
   };
 
   public constructor(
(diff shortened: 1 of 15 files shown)


src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
```diff
@@ -24,9 +28,17 @@ import { PricingDefinitionDialogComponent } from './pricing-definition/pricing-d
...
+  private viewAllAction = new TablePricingOverviewAction(this.viewingAllComponents).getActionDef();
   private createAction = new TableCreatePricingDefinitionAction().getActionDef();
   private editAction = new TableEditPricingDefinitionAction().getActionDef();
   private deleteAction = new TableDeletePricingDefinitionAction().getActionDef();
+  private viewAction = new TableViewPricingDefinitionsAction().getActionDef();
+  private defaultContext = {
+    entityID: '',
+    entityType: '',
+    entityName: ''
+  };
```
ClaudeROSSI, line 41: Why don't you put null as default values. Please avoid empty strings ''

src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
```diff
@@ -48,15 +60,17 @@ export class PricingDefinitionsTableDataSource extends DialogTableDataSource<Pri
...
     } else {
-      this.context.entityID = entityID;
-      this.context.entityType = entityType;
+      this.context.entityID = this.defaultContext.entityID = entityID;
+      this.context.entityType = this.defaultContext.entityType = entityType;
     }
-    this.context.entityName = entityName;
+    this.context.entityName = this.defaultContext.entityName = entityName;
+    //Force a table refresh
+    this.initDataSource(true);
```
ClaudeROSSI, line 73: can't we just do a refresh - like the refresh button 
c.f.:  this.refreshData.bind(this));

Also consider the context and set "withEntityInformation" to true when calling the endpoint to make sure you fetch the entity name as well.

src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
```diff
@@ -24,9 +28,17 @@ import { PricingDefinitionDialogComponent } from './pricing-definition/pricing-d
 
 @Injectable()
 export class PricingDefinitionsTableDataSource extends DialogTableDataSource<PricingDefinition> {
+  private viewingAllComponents = false;
+  private viewAllAction = new TablePricingOverviewAction(this.viewingAllComponents).getActionDef();
   private createAction = new TableCreatePricingDefinitionAction().getActionDef();
   private editAction = new TableEditPricingDefinitionAction().getActionDef();
   private deleteAction = new TableDeletePricingDefinitionAction().getActionDef();
+  private viewAction = new TableViewPricingDefinitionsAction().getActionDef();
```
ClaudeROSSI, line 36: ???? 

why do  we have a plural here 

private viewAction = new TableViewPricingDefinition**s**Action().getActionDef();

Please use the regular view action (with a eye as icon)

src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
```diff
@@ -139,26 +171,34 @@ export class PricingDefinitionsTableDataSource extends DialogTableDataSource<Pri
...
   }
 
   public buildTableActionsDef(): TableActionDef[] {
     const tableActionsDef = super.buildTableActionsDef();
-    return [
+    const actions: TableActionDef[] = [
       this.createAction,
-      ...tableActionsDef,
     ];
+    if(this.defaultContext.entityType === PricingEntity.TENANT){
```
ClaudeROSSI, line 184: missing space

src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
```diff
@@ -184,6 +244,13 @@ export class PricingDefinitionsTableDataSource extends DialogTableDataSource<Pri
             { dialogData: { ...pricingDefinition, context: this.context } }, this.refreshData.bind(this));
         }
         break;
+      case PricingButtonAction.VIEW_PRICING_DEFINITIONS:
+        if (actionDef.action) {
+          (actionDef as TableViewPricingDefinitionsActionDef).action(
+            PricingDefinitionDialogComponent, this.dialog,
+            { dialogData: { ...pricingDefinition, context: this.context }, dialogMode: DialogMode.VIEW }, this.refreshData.bind(this));
```
ClaudeROSSI, line 251: The DialogMode.VIEW seems correect here. But the result is not normal.
The dialog is disabled while it should be read-only.

disable (light gray) means data is inconsistent, inaccurate or meaningless
read-only (not editable) means data is correct but cannot be changed.

We need to find a solution. But this is not just for you ... this is a general issue in all UI(s)

ClaudeROSSI (changes requested):
some remarks

Thanks,
Claude

ClaudeROSSI:
@cabch : I asked Rohan to change the pricing definition list to allow showing all pricing definitions (in the pricing settings).
Please help me to review it.

PR #3373, opened 2022-09-21
added sortable true to public in chargingstations list

cabch:


cabch:
@KinzaBS sortable columns for this ui are : "name", "vendor" and "model" + "public" now. 
Are there other field you want to be sortable before we can merge this modification ?

ClaudeROSSI:
> Hi caro,
> 
> this is perfect, customers wanted also to sort in status but i don't think it's possible for performance reasons.
> 
> Best regards, Kinza

I'll merge that one.
Let's discuss if we need more filtering or a dedicated tab for OFFLINE stations.

ClaudeROSSI (approved):

PR #3377, opened 2022-09-25
Refactored Google Map

LucasBrazi06:


LucasBrazi06:
@ClaudeROSSI I introduced in this PR how the data passed to the dialogs should be typed as well as the results.
Types are in Dialog.ts file.
If you have someone that could type all the dialogs ;-)

cabch (approved):
sounds like it doesn't break the map dialog, ok for me

PR #3378, opened 2022-09-26
Npm install without force

LucasBrazi06:


LucasBrazi06:
@ClaudeROSSI Probably has to regenerate the lock file with an AMD64 Mac ;-)

PR #3383, opened 2022-09-28
Adding entity restrictions in details table

RohanAntony:
- Verify if a new column named restrictions is visible on the pricing table under integrations as shown below 
- Verify that it shows 1 to 5 icons as described below
 1. Connector Type: Title shows the type and connector power
 2. Hourglass: Title shows max and min duration for the definition
 3. Electric bolt: Title shows max and min energy for the definition
 4. Timer: Title shows the time duration for which the pricing is valid
 5. Daterange: Title shows the days of the week when the pricing is valid
- Verify that the title must be translated to respective languages
- Verify that the icon is hidden when the respective definition hasn't specified any constraint   
image


src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
```diff
@@ -126,6 +127,14 @@ export class PricingDefinitionsTableDataSource extends DialogTableDataSource<Pri
         headerClass: 'col-15p',
         class: 'col-15p',
         visible: this.viewingAllComponents,
+      },
```
ClaudeROSSI, line 130: @RohanAntony : I think this column should always be visible (even when "View all" is not ON).

ClaudeROSSI (commented):
just a comment

src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
RohanAntony: - Making **EntityType** visible when View All is off doesn't make sense as all entities listed will be of the same type. 
- For **EntityName** and **Restrictions**, I am adding the necessary changes.

ClaudeROSSI: I was talking about the restriction column only ;)

PR #3388, opened 2022-09-30
Refactored and simplified Tenant's components tab

LucasBrazi06:
Simplified tenant's components view
Increased contrast in lists
Fixed scrollbar is shown in tabs when the window is too small
Fixed scrollbar in side menu is always displayed
Small refactoring for the future dark theme
Fixed deprecated usage of ~ in SCSS
Fixed deprecated RxJs calls


LucasBrazi06:
@ClaudeROSSI Clean up of tenant view.
Goal: I will provide feature toggle in a new tab.

PR #3414, opened 2022-10-13
Do not build table when OnlyRecord flag is up

mahdibenromdhane:


Ramzay:
@mahdibenromdhane made a more generic fix to handle calls with only requestNumberOfRecords, you can check the commit here [88c50dd](https://github.com/sap-labs-france/ev-dashboard/commit/88c50dd89ae5a5597c825f3d0abf0c653e65d78f)

PR #3452, opened 2022-11-07
Fixed Admin can be set as Site Admin

LucasBrazi06:
Pass the User to Site Admin/Owner component via additional params in columnDef
Removed retrieving the user once per line from the backend in UserSites list
Added 'disabled' prop to table column def to have columns in read-only visible
Renamed Site Admin/Owner components


LucasBrazi06:
@Ramzay, my friendly piece of advices!

canUpdateUserSites should belong to the User and the dialog should open with this auth as parameter.
This will remove methods like 'buildDynamicTableDef' in the UserSitesTableDataSource and SiteUsersTableDataSource classes.
image

Add two new auth params canSetSiteAdmin and canSetSiteOwner to both siteUsers and userSites lists per record and move the code below from the template to the backend (the front should not do that, it should only use canXXX props!):
image

Last, ensure in the backend that canSetSiteAdmin and canSetSiteOwner auths are checked.
An Admin cannot be a Site Admin -> Throw exception.
In the end cleanup the code from table data sources (remove all *Dynamic* methods.)

Et arreter de vous faire des putains de noeuds dans la tete avec ces scenarios lol !!!

LucasBrazi06:
De rien 😘

src/app/pages/users/user-sites/user-sites-site-admin.component.ts
```diff
@@ -7,23 +7,25 @@ import { MessageService } from '../../../services/message.service';
...
 import { Utils } from '../../../utils/Utils';
 
 @Component({
   template: `
     <div class="d-flex justify-content-center">
       <mat-checkbox class="mx-auto"
-        [checked]="(row.siteAdmin ? row.siteAdmin : false)"
-        [disabled]="(!loggedUser.sitesAdmin.includes(row.site.id) && loggedUser.role !== 'A')"
+        [checked]="row?.siteAdmin || user?.role === UserRole.ADMIN"
+        [disabled]="(!isSiteAdmin && loggedUser.role !== UserRole.ADMIN) || columnDef.disabled || user?.role === UserRole.ADMIN"
```
cabch, line 18: Hi @LucasBrazi06, I don't understand why the Admin cannot be set as siteAdmin ?
Imagine siteAdmin (extended from basic) has new specific rights (eg: refund) the Admin won't have those new rights automatically, right ? They are not linked together as far as I know. How is it different from siteOwner ?

LucasBrazi06: @cabch Sorry for the late answer.
Site Admin for an Admin means nothing as he's already admin of all the Sites!
Site Admin is an elevation of privilege for Basic users on a given Site. 
Refund is only allowed to users with role Site Owner and an Admin can be Site Owner.
Voila, t'as compris ou je viens au Labs t'expliquer lol :-) !

cabch (approved):
ok

PR #3465, opened 2022-11-17
Migrate settings to abac

Ramzay:


src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
```diff
@@ -102,32 +112,28 @@ export class PricingDefinitionsTableDataSource extends DialogTableDataSource<Pri
...
+        class: 'col-15p',
+        visible: this.viewingAllComponents
+      },
+      {
+        id: 'entityName',
+        name: 'transactions.dialog.session.pricing_detail_entity_name',
+        headerClass: 'col-15p',
+        class: 'col-15p',
+        visible: this.viewingAllComponents
+      },
```
ClaudeROSSI, line 136: @Ramzay: This does not seem to work as expected.
The previous code was adding the entityType and entityName only when the "View all" mode.

The new way of doing it does not do the equivalent. The two columns are always visible.

image

ClaudeROSSI (changes requested):
@Ramzay : I have a side effect in the pricing list.

Check my comment.

src/app/shared/pricing-definitions/pricing-definitions-table-data-source.ts
Ramzay: @ClaudeROSSI, the "new way" was not working as expected, it always showed a column that was in the projectField array, even if in the table column definition we set visible = false.
I added a fix so that we can control the visibility of a column despite it being in the projected field array.

PR #3493, opened 2023-01-09
Show ocpi modal buttons on create mode

mahdibenromdhane:


Ramzay (changes requested):
@mahdibenromdhane your modifications bypasses authorizations, which is not what we want, anyone could generate the token if it's empty... You should look at an object "DialogParamsWithAuth" to pass authorizations to a dialog.

@cabch will make the fix.

Thanks

cabch:
sounds ok for me now, added cans at page loading + pass them to the dialog + change the ngIf condition

cabch:
deleted old verification to "currentEndpoint" as the new cans are global
also updated imports

Ramzay (approved):

PR #3523, opened 2023-03-14
Migrate statistics to abac

Ramzay:


ClaudeROSSI:
@AlixH : Ramzi has updated the statistics API regarding the ABAC authorizations. The endpoint is as usual backward-compatible with en current mobile app. But let's check it!
BTW: We also need to adapt the mobile app. to benefit from the new authorizations. Let's sync on this!

AlixH:
> @AlixH : Ramzi has updated the statistics API regarding the ABAC authorizations. The endpoint is as usual backward-compatible with en current mobile app. But let's check it! BTW: We also need to adapt the mobile app. to benefit from the new authorizations. Let's sync on this!

I can still see the curves as both admin and basic on the current public app

